PR #4, opened 2023-05-20
Faster duplicate check adblock-lean

Wizballs:
Same command, but approx 33% faster


lynxthecat:
Didn't quite work out?

Wizballs:
Yeah didn't work out so deleted that pull requested.  There should be
another pull request tho.

PR #8, opened 2023-05-28
Save some CPU cycles

Wizballs:
What do you think,
instead of cat blocklist.parts > blocklist
sed clean whitespace and format as local=/.../ >> blocklist


lynxthecat:
Looks good to me. I amended the commit message(s) to be more descriptive and made a few minor tweaks in respect of the logging and commenting.

PR #9, opened 2023-06-03
line count tweaks, rouge check simplification

Wizballs:
changed to grep for line counts.  Is slightly faster and also doesn't need to be piped to | wc -l
Rouge element only checking for local=/.../  .  Remove address=/.../ , server=/.../ , and .../.../# as not used by lean, and increases speed.
Commented out redundant good_line_count operation
Use sed '$a\' to add newline at end of each blocklist part including user blocklist.  Instead of printf

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -90,7 +90,7 @@ generate_preprocessed_blocklist_file()
 
 	rm -f /tmp/wget_err
 
-	preprocessed_blocklist_line_count=$(sed '/^\s*$/d' /tmp/blocklist | wc -l)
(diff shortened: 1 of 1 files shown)


lynxthecat:
Seems positive. Is this still using the busybox grep?

Wizballs:
Yes of course, whatever is built into Openwrt.  No extra packages needed

lynxthecat:
Superb. Let me run a test and I'll squash and merge the changes in.

lynxthecat:
The good line count is needed for status() because that is called independently from start(). So I reintroduced that. Otherwise I tested and all works fine. So I squashed and merged in. Nice work!

PR #10, opened 2023-06-03
Explicitly query localhost for nslookup

dave14305:
It is possible some users might disable the router’s own use of dnsmasq by setting localuse option to 0.

If /etc/resolv.conf does not point to 127.0.0.1 or ::1, nslookup will not be testing the local dnsmasq instance.


lynxthecat:
Looks good to me. Thanks @dave14305.

PR #21, opened 2024-02-17
Test scripts

daleq:
Test scripts to show system info (sw/ram/disk), run adblock-lean while simultaneously running vmstat to gather performance data.


lynxthecat:
I like the idea @daleq, but I would first like to experiment with alternative monitoring techniques.

PR #30, opened 2024-07-11
Add max_retries to config, avoid hardcoding it

friendly-bits:
- Better to set max_retries in config rather than hardcode it.
- This PR also implements retry loops without calling external binary (seq).

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -88,6 +88,7 @@ load_config()
 	fi
 
 	if [[ \
+	-z "${max_download_retries+set}" || \
 	-z "${blocklist_urls+set}" || \
 	-z "${allowlist_urls+set}" || \
 	-z "${local_allowlist_path+set}" || \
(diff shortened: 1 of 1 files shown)


lynxthecat:
How about we make it max_download_retries and then we should add it to the checks here:

https://github.com/lynxthecat/adblock-lean/pull/30/files#diff-759db58e0b7adb9f787c6c7b54caff784f1aa99d758c7e458c7d7768f83adb7aR105

friendly-bits:
> How about we make it max_download_retries and then we should add it to the checks here

done

friendly-bits:
Now actually done properly (I think)

lynxthecat:
Looks good.  Please can we change commit message to: 

Add max_download_retries to config file

to keep within the 50 character limit?

friendly-bits:
> Looks good. Please can we change commit message to:
> 
> Add max_download_retries to config file
> 
> to keep within the 50 character limit?

Done

PR #31, opened 2024-07-11
Use unique filenames, don't overwrite conf-script

friendly-bits:
- Avoids possible filename collisions with files generated by other software
- Avoids hardcoding paths
- Makes it easier for the user to figure out which software generated the files
- Avoids overwriting dnsmasq conf-script in order to allow other software to use that file
- Changes some calls to printf to follow best practice (https://www.shellcheck.net/wiki/SC2059)

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -396,11 +396,16 @@ generate_and_process_blocklist_file()
 
 	rm -f /tmp/dnsmasq_err
 
-	for blocklist_file_part_gz in /tmp/blocklist.*.gz
-	do
-		gunzip -c "${blocklist_file_part_gz}" | tee >(dnsmasq_test_output=$(dnsmasq --test -C - 2>&1); [[ $? != 0 ]] && "printf $dnsmasq_test_output" > /tmp/dnsmasq_err)
-		rm -f "${blocklist_file_part_gz}"
-	done | sort -u > /tmp/blocklist
+	{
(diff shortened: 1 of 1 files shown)


lynxthecat:
I am not convinced by introducing all the variables for the file paths. I am inclined instead just to use a /var/run/adblock-lean directory and hardcode all the paths there. 

I am convinced by the conf-script changes however. Could you raise a separate pull request for that (relative to the latest version of 'adblock-lean' on the 'improve-process-flow' branch)?

By the way, I simply put the lines of the init_awk_cmd function into start(), since this seemed to me to be just a function for the sake of a function.

friendly-bits:
> I am inclined instead just to use a /var/run/adblock-lean directory and hardcode all the paths there

Sure, this would be a functional solution. Not hard-coding paths is simply a matter of good practice but not a crucial thing.

> By the way, I simply put the lines of the init_awk_cmd function into start(), since this seemed to me to be just a function for the sake of a function.

There is no harm in separating parts of code which have a specific... well, function, in a separate subroutine. There is a benefit, though: it helps to keep code clean and readable. In this case, it helps to keep the start() function clean and readable. But it's your project and your choice how to code it.

P.s. sorry, somehow unneeded commits history slipped into this PR.

lynxthecat:
> P.s. sorry, somehow unneeded commits history slipped into this PR.

Not at all that is my fault for having rewritten GitHub history!

PR #32, opened 2024-07-13
Pipe downloads into further processing chains

friendly-bits:
- Reduce processing time by starting to process data as it's getting downloaded
- Fix bug where if rogue_element_action is "IGNORE", gzip will receive an empty string rather than the blocklist

NOTE: more testing is recommended, especially testing under fault conditions (eg failed downloads, rogue elements, excessive file size or insufficient lines count)

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -232,49 +232,58 @@ generate_preprocessed_blocklist_file_parts()
 		else
 			log_msg "No lines remaining in allowlist file part after sanitization."
 		fi
-		allowlist_line_count=$(( allowlist_line_count + allowlist_file_part_line_count))
+		allowlist_line_count=$(( allowlist_line_count + allowlist_file_part_line_count ))
 	else
 		log_msg "No local allowlist identified."
 	fi
 
 	for allowlist_url in ${allowlist_urls}
 	do
-		retry=0
+		retry=1
 		while [[ "$retry" -le "$max_download_retries" ]]
 		do
 			retry=$((retry + 1))
-			log_msg "Downloading new allowlist file part from: ${allowlist_url}."
-			uclient-fetch "${allowlist_url}" -O- --timeout=2 2> /tmp/uclient-fetch_err > /tmp/allowlist.${allowlist_id}
-			if grep -q "Download completed" /tmp/uclient-fetch_err
+			log_msg "Downloading and sanitizing new allowlist file part from: ${allowlist_url}."
+			uclient-fetch "${allowlist_url}" -O- --timeout=2 2> /tmp/uclient-fetch_err |
+			tee >(wc -c > /tmp/allowlist_file_part_size_B) |
+			tr 'A-Z' 'a-z' | sed 's/#.*$//; s/^[ \t]*//; s/[ \t]*$//; /^$/d; $a\' |
+			tee >(wc -l > /tmp/allowlist_file_part_line_count) > \
+				/tmp/allowlist.${allowlist_id}
+
+			allowlist_file_part_size_KB=$(( ($(cat /tmp/allowlist_file_part_size_B 2>/dev/null) + 0) / 1024 ))
+			allowlist_file_part_line_count="$(cat /tmp/allowlist_file_part_line_count 2>/dev/null)"
+			: "${allowlist_file_part_line_count:=0}"
+			rm -f /tmp/allowlist_file_part_size_B /tmp/allowlist_file_part_line_count
+
+			if ! grep -q "Download completed" /tmp/uclient-fetch_err
 			then
-				allowlist_file_part_size_KB=$(get_file_size_KB /tmp/allowlist.${allowlist_id})
-				log_msg "Download of new allowlist file part from: ${allowlist_url} succeeded (downloaded file size: ${allowlist_file_part_size_KB} KB)."
-				log_msg "Sanitizing allowlist file part."
-				# 1 Convert to lowercase; 2 Remove comment lines and trailing comments; 3 Remove trailing address hash, and all whitespace; 4 Add newline
-				allowlist_file_part_line_count="$(tr 'A-Z' 'a-z' < /tmp/allowlist.${allowlist_id} | sed 's/#.*$//; s/^[ \t]*//; s/[ \t]*$//; /^$/d; $a\' | tee -a /tmp/allowlist | wc -l)"
-				if [[ "${allowlist_file_part_line_count}" -gt 0 ]]
-				then
-					log_msg "Sanitized allowlist file part line count: ${allowlist_file_part_line_count}."
-				else
-					log_msg "No lines remaining in allowlist file part after sanitization."
-				fi
-				allowlist_line_count=$(( allowlist_line_count + allowlist_file_part_line_count))
-				rm -f  /tmp/allowlist.${allowlist_id}
-				allowlist_id=$((allowlist_id+1))
-				continue 2
-			else
 				log_msg "Download of new allowlist file part from: ${allowlist_url} failed."
 				log_msg "Sleeping for 5 seconds after failed download attempt."
 				sleep 5
 				continue
 			fi
+
+			log_msg "Download of new allowlist file part from: ${allowlist_url} succeeded (downloaded file size: ${allowlist_file_part_size_KB} KB)."
+
+			if [[ "${allowlist_file_part_line_count}" -gt 0 ]]
+			then
+				log_msg "Sanitized allowlist file part line count: ${allowlist_file_part_line_count}."
+				allowlist_line_count=$(( allowlist_line_count + allowlist_file_part_line_count ))
+				cat "/tmp/allowlist.${allowlist_id}" >> /tmp/allowlist
+			else
+				log_msg "No lines remaining in allowlist file part after sanitization."
+			fi
+
+			rm -f  "/tmp/allowlist.${allowlist_id}"
+			allowlist_id=$(( allowlist_id + 1 ))
+			continue 2
 		done
 		log_msg "Download failed after three failed download attempts. Continuing further operation."
 	done
 
 	rm -f /tmp/allowlist.*
 
-	if [[ -f "/tmp/allowlist" ]] && [[ $allowlist_line_count -gt 0 ]]
+	if [[ -f /tmp/allowlist ]] && [[ $allowlist_line_count -gt 0 ]]
 	then
 		log_msg "Successfully generated allowlist with ${allowlist_line_count} lines."
 		log_msg "Will remove any (sub)domain matches present in the generated allowlist from the blocklist file part(s) and append corresponding server entries to the combined blocklist."
@@ -294,84 +303,89 @@ generate_preprocessed_blocklist_file_parts()
 	else
 		log_msg "No local blocklist identified."
 	fi
-	
+
 	[[ -n "${blocklist_urls}" ]] && log_msg "Downloading new blocklist file part(s)."
 
 	preprocessed_blocklist_line_count=0
 	blocklist_id=1
 	for blocklist_url in ${blocklist_urls}
 	do
-		retry=0
+		retry=1
 		while [[ "$retry" -le "$max_download_retries" ]]
 		do
 			retry=$((retry + 1))
-			log_msg "Downloading new blocklist file part from: ${blocklist_url}."
-			uclient-fetch "${blocklist_url}" -O- --timeout=2 2> /tmp/uclient-fetch_err | head -c "${max_blocklist_file_part_size_KB}k" > "/tmp/blocklist.${blocklist_id}"
-			if grep -q "Download completed" /tmp/uclient-fetch_err
+			log_msg "Downloading, checking and sanitizing new blocklist file part from: ${blocklist_url}."
+			uclient-fetch "${blocklist_url}" -O- --timeout=2 2> /tmp/uclient-fetch_err | head -c "${max_blocklist_file_part_size_KB}k" |
+			tee >(wc -c > /tmp/blocklist_part_size_B) |
+			# 1 Convert to lowercase; 2 Remove comment lines and trailing comments; 3 Remove trailing address hash, and all whitespace; 4 Convert to local=; 5 Add newline
+			tr 'A-Z' 'a-z' | sed 's/#.*$//; s/^[ \t]*//; s/[ \t]*$//; /^$/d; s/^\(address=\|server=\)/local=/; $a\' |
+			if [[ "${use_allowlist}" == 1 ]]
 			then
-				blocklist_file_part_size_KB=$(get_file_size_KB /tmp/blocklist.${blocklist_id})
-				blocklist_file_part_line_count=$(grep -vEc '^\s*$|^#' /tmp/blocklist.${blocklist_id})
-
-				if [[ "${blocklist_file_part_line_count}" -ge "${min_blocklist_file_part_line_count}" ]]
-				then
-					log_msg "Download of new blocklist file part from: ${blocklist_url} succeeded (downloaded file size: ${blocklist_file_part_size_KB} KB; line count: ${blocklist_file_part_line_count})."
-
-					log_msg "Sanitizing blocklist file part."
-
-					# 1 Convert to lowercase; 2 Remove comment lines and trailing comments; 3 Remove trailing address hash, and all whitespace; 4 Convert to local=; 5 Add newline
-					tr 'A-Z' 'a-z' < /tmp/blocklist.${blocklist_id} | sed 's/#.*$//; s/^[ \t]*//; s/[ \t]*$//; /^$/d; s/^\(address=\|server=\)/local=/; $a\' |
-					if [[ "${use_allowlist}" == 1 ]]
-					then
-						${awk_cmd} -F'/' 'NR==FNR { allow[$0]; next } { n=split($2,arr,"."); addr = arr[n]; for ( i=n-1; i>=1; i-- ) { addr = arr[i] "." addr; if ( addr in allow ) next } } 1' "/tmp/allowlist" -
-					else
-						cat
-					fi |
-
-					if [[ "${rogue_element_action}" != "IGNORE" ]] 
-					then
-						log_msg "Checking for any rogue elements."
-						tee >(grep -Evn -m1 '^(local=/[[:alnum:]*][[:alnum:]*_.-]+/|bogus-nxdomain=[0-9.]+){0,1}$' > /tmp/rogue_element)
-					fi | gzip > "/tmp/blocklist.${blocklist_id}.gz"
-
-					if read -r rogue_element < /tmp/rogue_element
-					then
-						log_msg "Rogue element: '${rogue_element}' identified originating in blocklist file part from: ${blocklist_url}."
-
-						if [[ "${rogue_element_action}" == "STOP" ]]
-						then
-							return 1
-						else
-							log_msg "Skipping file part and continuing."
-							rm -f "/tmp/blocklist.${blocklist_id}" /tmp/rogue_element
-							continue 2
-						fi
-					fi
-
-					rm -f "/tmp/blocklist.${blocklist_id}" /tmp/rogue_element
-
-					preprocessed_blocklist_line_count=$(( preprocessed_blocklist_line_count + blocklist_file_part_line_count ))
-					blocklist_id=$((blocklist_id+1))
-					continue 2
-				else
-					log_msg "Downloaded blocklist file part line count: ${blocklist_file_part_line_count} less than configured minimum: ${min_blocklist_file_part_line_count}."
-					break
-				fi
+				${awk_cmd} -F'/' 'NR==FNR { allow[$0]; next } { n=split($2,arr,"."); addr = arr[n]; for ( i=n-1; i>=1; i-- ) { addr = arr[i] "." addr; if ( addr in allow ) next } } 1' /tmp/allowlist -
 			else
+				cat
+			fi |
+			tee >(wc -l > /tmp/blocklist_part_line_count) |
+
+			if [[ "${rogue_element_action}" != "IGNORE" ]]
+			then
+				tee >(sed -nE '\~^(local=/[[:alnum:]*][[:alnum:]*_.-]+/$|bogus-nxdomain=[0-9.]+$|$)~d;p;:1 n;b1' > /tmp/rogue_element)
+			else
+				cat
+			fi | gzip > /tmp/blocklist.${blocklist_id}.gz
+
+			blocklist_file_part_size_KB=$(( ($(cat /tmp/blocklist_part_size_B 2>/dev/null) + 0) / 1024 ))
+			blocklist_file_part_line_count="$(cat /tmp/blocklist_part_line_count 2>/dev/null)"
+			: "${blocklist_file_part_line_count:=0}"
+
+			rm -f /tmp/blocklist_part_size_B /tmp/blocklist_part_line_count
+
+			if ! grep -q "Download completed" /tmp/uclient-fetch_err
+			then
 				log_msg "Download of new blocklist file part from: ${blocklist_url} failed."
-				if [[ -f "/tmp/blocklist.${blocklist_id}" ]]
+				if [[ "${blocklist_file_part_size_KB}" -ge "${max_blocklist_file_part_size_KB}" ]]
 				then
-					blocklist_file_part_size_KB=$(get_file_size_KB /tmp/blocklist.${blocklist_id})
-					if [[ "${blocklist_file_part_size_KB}" -eq "${max_blocklist_file_part_size_KB}" ]]
-					then
 						log_msg "Downloaded blocklist file part size exceeded the maximum value set in config (${max_blocklist_file_part_size_KB} KB)."
 						log_msg "Consider either increasing this value in the config or removing the correasponding blocklist url."
 						break
-					fi
+				fi
+				log_msg "Sleeping for 5 seconds after failed download attempt."
+				sleep 5
+				continue
+			fi
+
+			if read -r rogue_element < /tmp/rogue_element
+			then
+				log_msg "Rogue element: '${rogue_element}' identified originating in blocklist file part from: ${blocklist_url}."
+
+				if [[ "${rogue_element_action}" == "STOP" ]]
+				then
+					return 1
+				else
+					log_msg "Skipping file part and continuing."
+					rm -f /tmp/rogue_element
+					continue 2
 				fi
 			fi
+
+			if [[ "${blocklist_file_part_line_count}" -ge "${min_blocklist_file_part_line_count}" ]]
+			then
+				log_msg "Download of new blocklist file part from: ${blocklist_url} succeeded (downloaded file size: ${blocklist_file_part_size_KB} KB; sanitized line count: ${blocklist_file_part_line_count})."
+
+				rm -f /tmp/rogue_element
+
+				preprocessed_blocklist_line_count=$(( preprocessed_blocklist_line_count + blocklist_file_part_line_count ))
+				blocklist_id=$((blocklist_id+1))
+				continue 2
+			else
+				log_msg "Downloaded blocklist file part line count: ${blocklist_file_part_line_count} less than configured minimum: ${min_blocklist_file_part_line_count}."
+				break
+			fi
+
 			log_msg "Sleeping for 5 seconds after failed download attempt."
 			sleep 5
 		done
+
 		if [[ "${download_failed_action}" == "STOP" ]]
 		then
 			log_msg "Exiting after three failed download attempts."


adblock-lean
```diff
@@ -306,72 +311,77 @@ generate_preprocessed_blocklist_file_parts()
...
+				rm -f /tmp/rogue_element
+
+				preprocessed_blocklist_line_count=$(( preprocessed_blocklist_line_count + blocklist_file_part_line_count ))
+				blocklist_id=$((blocklist_id+1))
+				continue 2
+			else
+				log_msg "Downloaded blocklist file part line count: ${blocklist_file_part_line_count} less than configured minimum: ${min_blocklist_file_part_line_count}."
+				break
+			fi
+			sleep_before_retry
```
lynxthecat, line 382: Let’s just expand this out without function.

friendly-bits: > Let’s just expand this out without function

The code in that function (sleep_before_retry) is called twice in this PR. Currently only the first call is strictly necessary because if everything works as intended, the second call will never get executed. It is in place because otherwise, if some of the logic is changed at a later point and execution is allowed to reach the end of the while loop, you will have a bug where download is retried without a delay.

So in this instance, the function is there to reduce code duplication. Do you still want to replace calls to it with duplicate code?

lynxthecat: Yes. We can also from the calls to safeguard against failure in future. I’d rather redesign the logic entirely than build in safeguards against not properly following the logic in the future.

friendly-bits: Sure, I will change this. However I'd like to give you an example of this safeguard applied in the wild to demonstrate that it's not far-fetched. The example is nftables and iptables chain policies. You can not create a base chain without a default policy. Why - because humans make mistakes, and a good coder recognizes this fact. If a subtle mistake may lead to software going into undetermined state then a good practice is to put safeguards in place to prevent that. I won't insist on this safeguard here but I wanted to explain the logic behind it and the fact that it's standard practice.

friendly-bits: > Yes. We can also remove the repetition to safeguard against failure in future.

Done.

lynxthecat: Thanks. Should be able to test this shortly.

adblock-lean
```diff
@@ -306,72 +306,77 @@ generate_preprocessed_blocklist_file_parts()
...
+					rm -f /tmp/rogue_element
+					continue 2
+				fi
+			fi
+
+			if [[ "${blocklist_file_part_line_count}" -ge "${min_blocklist_file_part_line_count}" ]]
+			then
+				log_msg "Download of new blocklist file part from: ${blocklist_url} succeeded (downloaded file size: ${blocklist_file_part_size_KB} KB; line count: ${blocklist_file_part_line_count})."
+
+				log_msg "Sanitizing blocklist file part."
```
lynxthecat, line 367: This should be deleted.

adblock-lean
```diff
@@ -306,72 +306,77 @@ generate_preprocessed_blocklist_file_parts()
...
+				else
+					log_msg "Skipping file part and continuing."
+					rm -f /tmp/rogue_element
+					continue 2
+				fi
+			fi
+
+			if [[ "${blocklist_file_part_line_count}" -ge "${min_blocklist_file_part_line_count}" ]]
+			then
+				log_msg "Download of new blocklist file part from: ${blocklist_url} succeeded (downloaded file size: ${blocklist_file_part_size_KB} KB; line count: ${blocklist_file_part_line_count})."
```
lynxthecat, line 365:
```suggestion
				log_msg "Download of new blocklist file part from: ${blocklist_url} succeeded (downloaded file size: ${blocklist_file_part_size_KB} KB; sanitized line count: ${blocklist_file_part_line_count})."
```

adblock-lean
lynxthecat, line 367: I think this line should be deleted.

adblock-lean
```diff
@@ -306,72 +306,77 @@ generate_preprocessed_blocklist_file_parts()
 		do
 			retry=$((retry + 1))
 			log_msg "Downloading new blocklist file part from: ${blocklist_url}."
```
lynxthecat, line 308:
```suggestion
			log_msg "Downloading and sanitizing new blocklist file part from: ${blocklist_url}."
```

adblock-lean
lynxthecat, line 367:
```suggestion
```

adblock-lean
```diff
@@ -306,72 +306,77 @@ generate_preprocessed_blocklist_file_parts()
...
+					continue 2
+				fi
+			fi
+
+			if [[ "${blocklist_file_part_line_count}" -ge "${min_blocklist_file_part_line_count}" ]]
+			then
+				log_msg "Download of new blocklist file part from: ${blocklist_url} succeeded (downloaded file size: ${blocklist_file_part_size_KB} KB; line count: ${blocklist_file_part_line_count})."
+
+				log_msg "Sanitizing blocklist file part."
+
```
lynxthecat, line 368:
```suggestion
```

adblock-lean
```diff
@@ -306,72 +306,77 @@ generate_preprocessed_blocklist_file_parts()
...
-				else
-					log_msg "Downloaded blocklist file part line count: ${blocklist_file_part_line_count} less than configured minimum: ${min_blocklist_file_part_line_count}."
-					break
-				fi
+			if [[ "${rogue_element_action}" != "IGNORE" ]]
+			then
+				log_msg "Checking for any rogue elements."
+				tee >(grep -Evn -m1 '^(local=/[[:alnum:]*][[:alnum:]*_.-]+/|bogus-nxdomain=[0-9.]+){0,1}$' > /tmp/rogue_element)
 			else
+				cat
```
lynxthecat, line 326: Nice fix!

lynxthecat:
How about also piping the allowlist downloads:

https://github.com/lynxthecat/adblock-lean/pull/32/files#diff-759db58e0b7adb9f787c6c7b54caff784f1aa99d758c7e458c7d7768f83adb7aR247

friendly-bits:
> How about also piping the allowlist downloads into their further processing chain

Done, needs testing. Also please see my latest post in the forum thread.

Wizballs:
Hey guys,
With the file file download piped straight into tr | sed | awk etc, 1.6k allowlist:
Processing improved 10 seconds (from 1m 33s to 1m 23s)!

(I've been away camping for a couple days :) so thought I could do some test.

Looks to be an issue with the way rouge elements are handled, and the download will retry three times if one is found.  I just dropped a domains list / different format blocklist into my url list:

```
Sun Jul 14 17:37:21 2024 user.notice adblock-lean: Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Sun Jul 14 17:37:21 2024 user.notice adblock-lean: Checking for any rogue elements.
Sun Jul 14 17:37:24 2024 user.notice adblock-lean: Download of new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt failed.
Sun Jul 14 17:37:24 2024 user.notice adblock-lean: Sleeping for 5 seconds after failed download attempt.
Sun Jul 14 17:37:29 2024 user.notice adblock-lean: Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Sun Jul 14 17:37:29 2024 user.notice adblock-lean: Checking for any rogue elements.
Sun Jul 14 17:37:31 2024 user.notice adblock-lean: Download of new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt failed.
Sun Jul 14 17:37:31 2024 user.notice adblock-lean: Sleeping for 5 seconds after failed download attempt.
Sun Jul 14 17:37:36 2024 user.notice adblock-lean: Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Sun Jul 14 17:37:36 2024 user.notice adblock-lean: Checking for any rogue elements.
Sun Jul 14 17:37:39 2024 user.notice adblock-lean: Download of new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt failed.
Sun Jul 14 17:37:39 2024 user.notice adblock-lean: Sleeping for 5 seconds after failed download attempt.
Sun Jul 14 17:37:44 2024 user.notice adblock-lean: Skipping file part and continuing.
```

friendly-bits:
@Wizballs thanks for testing. Can you provide your config file contents? I'd like to see if I can reproduce the issue.

Wizballs:
Sure thing

I do have report failure / success variable to send an email on every run.  But excluded from here due to passwords etc.
It's this url that has rogue elements (well for the purpose of testing this anyway).

```
# adblock-lean configuration options

# One or more dnsmasq blocklist urls separated by spaces
blocklist_urls="https://raw.githubusercontent.com/hagezi/dns-blocklists/main/dnsmasq/tif.txt https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt https://raw.githubusercontent.com/sjhgvr/oisd/main/dnsmasq2_big.txt https://raw.githubusercontent.com/hagezi/dns-blocklists/main/dnsmasq/tif-ips.txt https://raw.githubusercontent.com/hagezi/dns-blocklists/main/dnsmasq/pro.txt"

# One or more allowlist urls separated by spaces
allowlist_urls="https://raw.githubusercontent.com/hagezi/dns-blocklists/main/wildcard/whitelist-referral-onlydomains.txt"
# allowlist_urls=""

# Path to optional local allowlist/blocklist files in the form:
# site1.com
# site2.com
local_allowlist_path="/root/adblock-lean/allowlist"
local_blocklist_path="/root/adblock-lean/blocklist"

# Mininum number of lines of any individual downloaded blocklist part
min_blocklist_file_part_line_count=1
# Maximum size of any individual downloaded blocklist part
max_blocklist_file_part_size_KB=30000
# Maximum total size of combined, processed blocklist
max_blocklist_file_size_KB=40000
# Minimum number of good lines in final postprocessed blocklist
min_good_line_count=100000

# compress blocklist to save memory once blocklist has been loaded
compress_blocklist=0 # enable (1) or disable (0) blocklist compression

# restart dnsmasq if previous blocklist was extracted and before generation of
# new blocklist thereby to free up memory during generaiton of new blocklist
initial_dnsmasq_restart=0 # enable (1) or disable (0) initial dnsmasq restart

# Rogue element action:
# SKIP_PARTIAL - skip failed blocklist file part and continue blocklist generation
# STOP - stop blocklist generation (and fallback to previous blocklist if available)
# IGNORE - ignore any rogue elements (warning: use with caution)
rogue_element_action="SKIP_PARTIAL"

# Maximum number of download retries
max_download_retries=3

# Download failed action:
# SKIP_PARTIAL - skip failed blocklist file part and continue blocklist generation
# STOP - stop blocklist generation (and fallback to previous blocklist if available)
download_failed_action="SKIP_PARTIAL"

# The following shell variables are invoked using:
# 'eval ${report_failure}' and 'eval ${report_success}'
# thereby to facilitate sending e.g. mailsend/sms notifications
# The variables '${failure_msg}' and '${success_msg}' can be employed
report_failure=""
report_success=""

# Start delay in seconds when service is started from system boot
boot_start_delay_s=120

```

lynxthecat:
Given the m1:

https://github.com/lynxthecat/adblock-lean/blob/81d79c7d21b90a8da936d1a5b5a1d56f238a8992/adblock-lean#L333

we need to either drop the m1 to allow download to finish or probably better swap check for download completed with check for rogue elements.

friendly-bits:
> we need to either drop the m1 to allow download to finish
That indeed "solves" the issue but I don't understand why. I assume that the issue happens when grep finds a rogue element and that somehow interrupts the pipe, so "Download completed" doesn't get written into /tmp/uclient-fetch_err. However when testing with another command, I'm not getting the same effect.
```
echo "1"$'\n'"2"$'\n'"3" > test
$ cat test | if [ true ]; then tee >(grep -m1 '#' > grepped); else cat; fi
1
2
3

$ cat grepped
1
```

So I don't understand how and why uclient-fetch gets the signal to stop.

> or probably better swap check for download completed with check for rogue elements.
This we can't do: the check for rogue elements is attached to the pipe while the check for download completed isn't.

lynxthecat:
https://github.com/lynxthecat/adblock-lean/blob/81d79c7d21b90a8da936d1a5b5a1d56f238a8992/adblock-lean#L358

friendly-bits:
All right, so the underlying issue is that for some reason, if the command called by the `tee >(<command>)` construct exits prematurely, uclient-fetch receives an interrupt which causes it to think that the connection timed out:
```
Downloading 'https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt'
Connecting to 185.199.108.133:443
Writing to stdout
-                     99% |****************************** |  3067k  0:00:00 ETAConnection error: Connection timed out
```

I gave up on trying to understand why but there is a workaround: use sed :)
`tee >(sed -nE '\~^(local=/[[:alnum:]*][[:alnum:]*_.-]+/$|bogus-nxdomain=[0-9.]+$|$)~d;{p;:1 n;b1}' > /tmp/rogue_element)`

What I'm trying to avoid is situation where you get a file with tons of rogue elements and then you write all that file for no reason (while you are only looking at the first match later anyway). So the above sed command only writes the 1st match, then quickly skips through the rest of the lines (rather than quitting like `grep -m1` does).

Wizballs:
Using the sed line above (replacing grep), I'm still seeing the same issue ie, three attempts to download. 

```
Sun Jul 14 20:55:27 2024 user.notice adblock-lean: Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Sun Jul 14 20:55:28 2024 user.notice adblock-lean: Checking for any rogue elements.
Sun Jul 14 20:55:30 2024 user.notice adblock-lean: Download of new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt failed.
Sun Jul 14 20:55:30 2024 user.notice adblock-lean: Sleeping for 5 seconds after failed download attempt.
Sun Jul 14 20:55:35 2024 user.notice adblock-lean: Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Sun Jul 14 20:55:35 2024 user.notice adblock-lean: Checking for any rogue elements.
Sun Jul 14 20:55:37 2024 user.notice adblock-lean: Download of new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt failed.
Sun Jul 14 20:55:37 2024 user.notice adblock-lean: Sleeping for 5 seconds after failed download attempt.
Sun Jul 14 20:55:42 2024 user.notice adblock-lean: Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Sun Jul 14 20:55:42 2024 user.notice adblock-lean: Checking for any rogue elements.
Sun Jul 14 20:55:45 2024 user.notice adblock-lean: Download of new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt failed.
Sun Jul 14 20:55:45 2024 user.notice adblock-lean: Sleeping for 5 seconds after failed download attempt.
Sun Jul 14 20:55:50 2024 user.notice adblock-lean: Skipping file part and continuing.
```

Wizballs:
> Also I don't think that there is a point in printing this line: "Checking for any rogue elements." It made sense when that check was a separate step, but now when it's attached to the pipe, there is no more separate step for this check.

There might be some value from an information perspective.  It separates the different processing components in the log.  Ie I can see (in seconds) how long it took to sanitize, and also how long to run the rogue check.

friendly-bits:
> Using the sed line above (replacing grep), I'm still seeing the same issue ie, three attempts to download.

That's odd. I double checked and I'm not getting these errors with the above command. Rather I'm getting this:
```
Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Checking for any rogue elements.
Rogue element: '0-02.net' identified originating in blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Skipping file part and continuing.
```
... as designed.

 Are you sure that you are running the updated code in your /etc/init.d/ ? The line to replace is line 333.

Wizballs:
Yes line 333 exactly:

Is there anything else that needs to be changed also?

```
then
				log_msg "Checking for any rogue elements."
				tee >(sed -nE '\~^(local=/[[:alnum:]*][[:alnum:]*_.-]+/$|bogus-nxdomain=[0-9.]+$|$)~d;{p;:1 n;b1}' > /tmp/rogue_element)
			else
```




> It doesn't work like this anymore. "Checking for any rogue elements" is now printed as soon as data starts flowing in the pipe. Should be about the same time as start of the sanitization.

Ahh yes I can see this now in the log.  Both downloading and sanitizing, and rogue check both have the same timestamp - thanks.

friendly-bits:
@Wizballs make sure you are using this code:

https://github.com/lynxthecat/adblock-lean/blob/81d79c7d21b90a8da936d1a5b5a1d56f238a8992/adblock-lean

And replace line 333. That's it, as far as I'm aware...

friendly-bits:
Nope, it was a technical issue on my part, the file in /etc/init.d didn't get updated. Now that it is, not getting the error anymore...

friendly-bits:
@Wizballs Here's the updated code with line 333 changed, just in case... Try to use that?
https://pastebin.com/iRGRFEdN

Wizballs:
Well strangely enough, even directly from pastebin I'm getting the same failed log?

I'm on Openwrt 23.05.3

@lynxthecat are you able to run also and see what happens?

friendly-bits:
@Wizballs and if you use the grep command but remove the `-m1` option?

friendly-bits:
@Wizballs please insert this line:
`set -x`
before this line:
`uclient-fetch "${blocklist_url}" -O- --timeout=2 2> /tmp/uclient-fetch_err | head -c "${max_blocklist_file_part_size_KB}k" |`

and this line:
`set +x`

**after** this one:
`fi | gzip > /tmp/blocklist.${blocklist_id}.gz`

then run `/etc/init.d/adblock-lean start` and post your output.
(this is the command you are running to get those errors, right?)

Wizballs:
grep with -m1 removed works

```
Sun Jul 14 22:29:54 2024 user.notice adblock-lean: Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Sun Jul 14 22:29:54 2024 user.notice adblock-lean: Checking for any rogue elements. grep remove -m1
Sun Jul 14 22:29:59 2024 user.notice adblock-lean: Rogue element: '1:[0-02.net](http://0-02.net/)' identified originating in blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Sun Jul 14 22:29:59 2024 user.notice adblock-lean: Skipping file part and continuing.
```

trying set +/- x as above next

Wizballs:
with set +/- x, the domain file still tried to download three times.  I can see there is a lot more verbose:

```
root@OpenWrt:~# service adblock-lean restart
Stopping adblock-lean.
Removing any adblock-lean blocklist files in /tmp/dnsmasq.d/ and restarting dnsmasq.
Removing any leftover adblock-lean temporary files.
Stopped adblock-lean.
Started adblock-lean.
gawk detected so using gawk for fast (sub)domain match removal.
No existing compressed or uncompressed blocklist identified.
No local allowlist identified.
Downloading and sanitizing new allowlist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/wildcard/whitelist-referral-onlydomains.txt.
Download of new allowlist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/wildcard/whitelist-referral-onlydomains.txt succeeded (downloaded file size: 30 KB).
Sanitized allowlist file part line count: 1569.
Successfully generated allowlist with 1569 lines.
Will remove any (sub)domain matches present in the generated allowlist from the blocklist file part(s) and append corresponding server entries to the combined blocklist.
No local blocklist identified.
Downloading new blocklist file part(s).
Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
+ uclient-fetch https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt -O- '--timeout=2'
+ gzip
+ + tee /dev/fd/64
tr A-Z a-z
+ tee /dev/fd/64
+ head -c 30000k
+ wc -c
+ sed 's/#.*$//; s/^[ \t]*//; s/[ \t]*$//; /^$/d; s/^\(address=\|server=\)/local=/; $a\'
+ '[[' 1 '==' 1 ]]
+ gawk -F/ 'NR==FNR { allow[$0]; next } { n=split($2,arr,"."); addr = arr[n]; for ( i=n-1; i>=1; i-- ) { addr = arr[i] "." addr; if ( addr in allow ) next } } 1' /tmp/allowlist -
+ '[[' SKIP_PARTIAL '!=' IGNORE ]]
+ log_msg 'Checking for any rogue elements.  sed +/- x'
+ local 'msg=Checking for any rogue elements.  sed +/- x'
+ printf 'Checking for any rogue elements.  sed +/- x\n'
Checking for any rogue elements.  sed +/- x
+ logger -t adblock-lean 'Checking for any rogue elements.  sed +/- x'
+ wc -l
+ tee /dev/fd/64
+ sed -nE '\~^(local=/[[:alnum:]*][[:alnum:]*_.-]+/$|bogus-nxdomain=[0-9.]+$|$)~d;{p;:1 n;b1}'
sed: can't find label for jump to '1}'
+ set +x
Download of new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt failed.
Sleeping for 5 seconds after failed download attempt.
Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
+ uclient-fetch https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt -O- '--timeout=2'
+ + gzip
'[[' SKIP_PARTIAL '!=' IGNORE ]]
+ log_msg 'Checking for any rogue elements.  sed +/- x'
+ local 'msg=Checking for any rogue elements.  sed +/- x'
+ + sed 's/#.*$//; s/^[ \t]*//; s/[ \t]*$//; /^$/d; s/^\(address=\|server=\)/local=/; $a\'
printf 'Checking for any rogue elements.  sed +/- x\n'
+ Checking for any rogue elements.  sed +/- x
+ logger -t adblock-lean 'Checking for any rogue elements.  sed +/- x'
tr A-Z a-z
+ tee /dev/fd/64
+ '[[' 1+  '==' 1 ]]
+ tee /dev/fd/64
gawk -F/ 'NR==FNR { allow[$0]; next } { n=split($2,arr,"."); addr = arr[n]; for ( i=n-1; i>=1; i-- ) { addr = arr[i] "." addr; if ( addr in allow ) next } } 1'+ head -c 30000k
 /tmp/allowlist -
+ wc -c
+ wc -l
+ tee /dev/fd/64
+ sed -nE '\~^(local=/[[:alnum:]*][[:alnum:]*_.-]+/$|bogus-nxdomain=[0-9.]+$|$)~d;{p;:1 n;b1}'
sed: can't find label for jump to '1}'
+ set +x
Download of new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt failed.
Sleeping for 5 seconds after failed download attempt.
Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
+ uclient-fetch https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt -O- '--timeout=2'
+ gzip
+ '[[' SKIP_PARTIAL '!=' IGNORE ]]
+ log_msg 'Checking for any rogue elements.  sed +/- x'
+ local 'msg=Checking for any rogue elements.  sed +/- x'
+ printf 'Checking for any rogue elements.  sed +/- x\n'
Checking for any rogue elements.  sed +/- x
+ logger -t adblock-lean 'Checking for any rogue elements.  sed +/- x'
+ + tr+  A-Z a-z
+ tee /dev/fd/64
sed 's/#.*$//; s/^[ \t]*//; s/[ \t]*$//; /^$/d; s/^\(address=\|server=\)/local=/; $a\''[['
 1+ tee /dev/fd/64 '==' 1 ]]
+ gawk -F/ 'NR==FNR { allow[$0]; next } { n=split($2,arr,"."); addr = arr[n]; for ( i=n-1; i>=1; i-- ) { addr = arr[i] "." addr; if ( addr in allow ) next } } 1' /tmp/allowlist -

+ head -c 30000k
+ wc -c
+ tee /dev/fd/64
+ sed -nE '\~^(local=/[[:alnum:]*][[:alnum:]*_.-]+/$|bogus-nxdomain=[0-9.]+$|$)~d;{p;:1 n;b1}'+ wc -l

sed: can't find label for jump to '1}'
+ set +x
Download of new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt failed.
Sleeping for 5 seconds after failed download attempt.
Skipping file part and continuing.
Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/dnsmasq/pro.txt.
+ uclient-fetch https://raw.githubusercontent.com/hagezi/dns-blocklists/main/dnsmasq/pro.txt -O- '--timeout=2'
+ gzip
+ '[[' SKIP_PARTIAL '!=' IGNORE ]]
+ log_msg 'Checking for any rogue elements.  sed +/- x'
+ tr A-Z a-z
+ + + local 'msg=Checking for any rogue elements.  sed +/- x'
+ printf 'Checking for any rogue elements.  sed +/- x\n'
Checking for any rogue elements.  sed +/- x
+ logger -t adblock-lean 'Checking for any rogue elements.  sed +/- x'
+ + tee /dev/fd/64
+ head -c 30000k
sed 's/#.*$//; s/^[ \t]*//; s/[ \t]*$//; /^$/d; s/^\(address=\|server=\)/local=/; $a\'
'[['+ wc -c
 1 '==' 1 ]]
+ gawk -F/ 'NR==FNR { allow[$0]; next } { n=split($2,arr,"."); addr = arr[n]; for ( i=n-1; i>=1; i-- ) { addr = arr[i] "." addr; if ( addr in allow ) next } } 1' /tmp/allowlist -
tee /dev/fd/64
+ wc -l
+ tee /dev/fd/64
+ sed -nE '\~^(local=/[[:alnum:]*][[:alnum:]*_.-]+/$|bogus-nxdomain=[0-9.]+$|$)~d;{p;:1 n;b1}'
+ set +x
Download of new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/dnsmasq/pro.txt succeeded (downloaded file size: 4269 KB; sanitized line count: 158657).
Successfully generated preprocessed blocklist file based on 158657 lines.
Performing dnsmasq --test on the blocklist file parts and sorting and merging the blocklist lines into a single blocklist file.
The dnsmasq --test on all of the blocklist file parts passed.
Processed blocklist file size: 4307 KB.
New blocklist file check passed.
Successfully imported new blocklist file for use by dnsmasq with size: 4307 KB.
Processing time for blocklist generation and import: 0m:32s.
Restarting dnsmasq.
Restart of dnsmasq completed.
Checking dnsmasq instance.
The dnsmasq check passed with new blocklist file.
New blocklist installed with good line count: 160225.
Mail sent successfully
The locally installed adblock-lean seems to be outdated.
Consider running: 'service adblock-lean update' to update it to the latest version.
```

friendly-bits:
> sed: can't find label for jump to '1}'

This is the culprit. The infamous sed versions syntax differences. I'll rewrite the command to be more universal...

friendly-bits:
@Wizballs please test with this command in line 333:
`tee >(sed -nE '\~^(local=/[[:alnum:]*][[:alnum:]*_.-]+/$|bogus-nxdomain=[0-9.]+$|$)~d;p;:1 n;b1' > /tmp/rogue_element)`

adblock-lean
```diff
@@ -301,77 +310,80 @@ generate_preprocessed_blocklist_file_parts()
...
+				rm -f /tmp/rogue_element
+
+				preprocessed_blocklist_line_count=$(( preprocessed_blocklist_line_count + blocklist_file_part_line_count ))
+				blocklist_id=$((blocklist_id+1))
+				continue 2
+			else
+				log_msg "Downloaded blocklist file part line count: ${blocklist_file_part_line_count} less than configured minimum: ${min_blocklist_file_part_line_count}."
+				break
 			fi
-			log_msg "Sleeping for 5 seconds after failed download attempt."
```
lynxthecat, line 372: Actually maybe we should keep these in just in case, as you suggested. But I still prefer to keep inline even with the duplication.

lynxthecat:
> https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt

It works:

```
root@OpenWrt-1:/tmp/adblock-lean# service adblock-lean start
Started adblock-lean.
gawk detected so using gawk for fast (sub)domain match removal.
Exporting and saving existing compressed blocklist.
Found local allowlist.
Sanitizing allowlist file part.
Sanitized allowlist file part line count: 2.
Successfully generated allowlist with 2 lines.
Will remove any (sub)domain matches present in the generated allowlist from the blocklist file part(s) and append corresponding server entries to the combined blocklist.
No local blocklist identified.
Downloading new blocklist file part(s).
Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Checking for any rogue elements.
Rogue element: '0-02.net' identified originating in blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Skipping file part and continuing.
Downloading and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/dnsmasq/pro.txt.
Checking for any rogue elements.
```

adblock-lean
```diff
@@ -232,49 +232,58 @@ generate_preprocessed_blocklist_file_parts()
...
+			log_msg "Downloading and sanitizing new allowlist file part from: ${allowlist_url}."
+			uclient-fetch "${allowlist_url}" -O- --timeout=2 2> /tmp/uclient-fetch_err |
+			tee >(wc -c > /tmp/allowlist_file_part_size_B) |
+			tr 'A-Z' 'a-z' | sed 's/#.*$//; s/^[ \t]*//; s/[ \t]*$//; /^$/d; $a\' |
+			tee >(wc -l > /tmp/allowlist_file_part_line_count) > \
+				/tmp/allowlist.${allowlist_id}
+
+			allowlist_file_part_size_KB=$(( ($(cat /tmp/allowlist_file_part_size_B 2>/dev/null) + 0) / 1024 ))
+			allowlist_file_part_line_count="$(cat /tmp/allowlist_file_part_line_count 2>/dev/null)"
+			: "${allowlist_file_part_line_count:=0}"
```
lynxthecat, line 255: What does this line do?

lynxthecat:
@friendly-bits I also have a question concerning the use of stream and -n with grep in respect of the rogue element check. The output is:

```
Rogue element: '0-02.net' identified originating in blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/domains/light.txt.
Skipping file part and continuing.
``` 

so no line number . Is that to be expected here given that we are using a stream? Not possible to retain line numbers in this way?

friendly-bits:
> so no line number. Any idea why?

Wrote this to @Wizballs above:

> Also I just realized that there is no point in printing the line number because this happens after sanitization, so that number won't correspond to the line number in the input file.
> 
> Also I don't think that there is a point in printing this line: "Checking for any rogue elements." It made sense when that check was a separate step, but now when it's attached to the pipe, there is no more separate step for this check.

lynxthecat:
Makes sense. How about we remove the "Checking for any rogue elements." message and instead modify the downloading message once more to specify:

```
Downloading, checking and sanitizing new blocklist file part from:
```

adblock-lean
```diff
@@ -294,84 +303,89 @@ generate_preprocessed_blocklist_file_parts()
...
 	do
-		retry=0
+		retry=1
 		while [[ "$retry" -le "$max_download_retries" ]]
 		do
 			retry=$((retry + 1))
-			log_msg "Downloading new blocklist file part from: ${blocklist_url}."
-			uclient-fetch "${blocklist_url}" -O- --timeout=2 2> /tmp/uclient-fetch_err | head -c "${max_blocklist_file_part_size_KB}k" > "/tmp/blocklist.${blocklist_id}"
-			if grep -q "Download completed" /tmp/uclient-fetch_err
+			log_msg "Downloading and sanitizing new blocklist file part from: ${blocklist_url}."
```
lynxthecat, line 317:
```suggestion
			log_msg "Downloading, checking and sanitizing new blocklist file part from: ${blocklist_url}."
```

friendly-bits:
@lynxthecat @Wizballs all done, please test and let me know.

friendly-bits:
Oops, changed the wrong download message previously - fixed now.

lynxthecat:
From my testing all seems to be working well now. @Wizballs? 

Nicely done @friendly-bits!

Wizballs:
Morning guys.  I can see merged already, but just reporting working great for me now also :)

PR #33, opened 2024-07-14
Use /var/run/ and trap to cleanup on exit

lynxthecat:
Switch run directory from /tmp/ to /var/run/adblock-lean and call `rm -rf /var/run/adblock-lean` on exit.

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -26,6 +26,17 @@ adblock-lean custom commands:
 	gen_config	generate default config
 	update		update adblock-lean to the latest version"
 
+mkdir -p /var/run/adblock-lean
+
+trap cleanup_and_exit INT TERM EXIT
+
+cleanup_and_exit()
+{
+	rm -rf /var/run/adblock-lean
+	trap - INT TERM EXIT
+	exit
+}
+
 get_file_size_KB()
 {
 	du -bk "$1" | awk '{print $1}'
@@ -216,7 +227,7 @@ check_blocklist_compression_support()
 
 generate_preprocessed_blocklist_file_parts()
 {
-	rm -f /tmp/allowlist*
+	rm -f /var/run/adblock-lean/allowlist*
 	allowlist_line_count=0
 	allowlist_id=0
 
@@ -225,7 +236,7 @@ generate_preprocessed_blocklist_file_parts()
 		log_msg "Found local allowlist."
 		log_msg "Sanitizing allowlist file part."
 		# 1 Convert to lowercase; 2 Remove comment lines and trailing comments; 3 Remove trailing address hash, and all whitespace; 4 Add newline
-		allowlist_file_part_line_count="$(tr 'A-Z' 'a-z' < "${local_allowlist_path}" | sed 's/#.*$//; s/^[ \t]*//; s/[ \t]*$//; /^$/d; $a\' | tee /tmp/allowlist | wc -l)"
+		allowlist_file_part_line_count="$(tr 'A-Z' 'a-z' < "${local_allowlist_path}" | sed 's/#.*$//; s/^[ \t]*//; s/[ \t]*$//; /^$/d; $a\' | tee /var/run/adblock-lean/allowlist | wc -l)"
 		if [[ "${allowlist_file_part_line_count}" -gt 0 ]]
 		then
 			log_msg "Sanitized allowlist file part line count: ${allowlist_file_part_line_count}."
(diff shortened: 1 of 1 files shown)


lynxthecat:
I haven't tested this yet, but @friendly-bits how does this look to you?

adblock-lean
```diff
@@ -495,16 +505,16 @@ restart_dnsmasq()
 
 export_existing_blocklist()
 {
-	if [[ -f /tmp/dnsmasq.d/.blocklist.gz ]]
+	if [[ -f /var/run/adblock-lean/dnsmasq.d/.blocklist.gz ]]
```
friendly-bits, line 508: does it need to be in subdirectory dnsmasq.d? or is this just an oversight?

adblock-lean
```diff
@@ -26,6 +26,16 @@ adblock-lean custom commands:
 	gen_config	generate default config
 	update		update adblock-lean to the latest version"
 
+mkdir -p /var/run/adblock-lean
+
+trap cleanup INT TERM EXIT
```
friendly-bits, line 31: I think you can also include QUIT and HUP. Don't ask me what each does :) People just do that :)

adblock-lean
```diff
@@ -26,6 +26,16 @@ adblock-lean custom commands:
...
+
+cleanup()
+{
+	log_msg "Cleaning up and exiting adblock-lean."
+	rm -rf /var/run/adblock-lean
```
friendly-bits, line 36: Why not clean all temp files here as well? When OOM crashes occur, some of them don't get cleaned up.

adblock-lean
```diff
@@ -529,25 +539,25 @@ restore_saved_blocklist()
...
 	if [[ "${compress_blocklist}" == 1 ]]
 	then
 		clean_dnsmasq_dir
-		printf "conf-script=\"busybox sh /tmp/dnsmasq.d/.extract_blocklist\"\n" > /tmp/dnsmasq.d/conf-script
-		printf "busybox gunzip -c /tmp/dnsmasq.d/.blocklist.gz\nexit 0\n" > /tmp/dnsmasq.d/.extract_blocklist
-		gzip -f /tmp/blocklist
-		mv /tmp/blocklist.gz /tmp/dnsmasq.d/.blocklist.gz
-		imported_blocklist_file_size_KB=$(get_file_size_KB /tmp/dnsmasq.d/.blocklist.gz)
+		printf "conf-script=\"busybox sh /var/run/adblock-lean/dnsmasq.d/.extract_blocklist\"\n" > /var/run/adblock-lean/dnsmasq.d/conf-script
+		printf "busybox gunzip -c /var/run/adblock-lean/dnsmasq.d/.blocklist.gz\nexit 0\n" > /var/run/adblock-lean/dnsmasq.d/.extract_blocklist
```
friendly-bits, line 553: I think dnsmasq looks for conf-script in /tmp/dnsmasq.d/ ? I'm still not really sure how this works but that's the impression I got.

adblock-lean
```diff
@@ -529,25 +539,25 @@ restore_saved_blocklist()
 
 clean_dnsmasq_dir()
 {
-	rm -f /tmp/dnsmasq.d/.blocklist.gz /tmp/dnsmasq.d/blocklist /tmp/dnsmasq.d/conf-script /tmp/dnsmasq.d/.extract_blocklist
+	rm -f /var/run/adblock-lean/dnsmasq.d/.blocklist.gz /var/run/adblock-lean/dnsmasq.d/blocklist /var/run/adblock-lean/dnsmasq.d/conf-script /var/run/adblock-lean/dnsmasq.d/.extract_blocklist
 }
```
friendly-bits, line 543: (another conf-script instance)

adblock-lean
lynxthecat: Oops! Good catch. Will fix.

adblock-lean
lynxthecat: Doesn’t removing the directory remove all the temporary files?

adblock-lean
lynxthecat: I need to fix this too!

adblock-lean
lynxthecat: Yes, thanks. Forgot about all the dnsmasq.d special case when I did the find and replace.

lynxthecat:
Thanks for the comments @friendly-bits. Will need to fix all the references to the dnsmasq.d directory.

adblock-lean
friendly-bits: My bad, I didn't realize this was a directory.

adblock-lean
friendly-bits, line 1: While we're at it, I think the config file belongs in /etc/adblock-lean. We shouldn't be touching /root.

lynxthecat: You are right. But I think that’s one for another day and maybe switching over to the whole OpenWrt UCI mechanism.

friendly-bits: BTW I wrote config management code for geoip-shell which doesn't rely on UCI. Not sure if this is relevant for adblock-lean but in case you will want an alternative to UCI, I have well-tested code for this.

friendly-bits: Ah, almost missed my chance to talk about hardcoding :) Here goes. If paths were not hard-coded in 1000 different instances, changing them would have been a much easier job with much lesser likelihood of creating bugs. Which is why avoiding hardcoding values (and especially paths) is considered good practice. Just sayin'.

PR #47, opened 2024-07-20
Improve error handling and delete obsolete files

friendly-bits:
- Fix rogue element check
- Always delete temp files before starting to download a new part.
- Always delete gzipped blocklist part if there is a problem with it
- If blocklist part line count is below minimum, retry download rather than giving up

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -346,6 +346,7 @@ generate_preprocessed_blocklist_file_parts()
 
 			if ! grep -q "Download completed" /var/run/adblock-lean/uclient-fetch_err
 			then
+				rm -f "/var/run/adblock-lean/allowlist.${allowlist_id}"
 				log_msg -err "Download of new allowlist file part from: ${allowlist_url} failed."
 				if [[ "${retry}" -lt "${max_download_retries}" ]]
 				then
@@ -404,6 +405,7 @@ generate_preprocessed_blocklist_file_parts()
 	blocklist_id=1
 	for blocklist_url in ${blocklist_urls}
 	do
+		rm -f /var/run/adblock-lean/rogue_element /var/run/adblock-lean/dnsmasq_err /var/run/adblock-lean/uclient-fetch_err
 		retry=0
 		while [[ "${retry}" -le "${max_download_retries}" ]]
 		do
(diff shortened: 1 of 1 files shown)


adblock-lean
```diff
@@ -437,15 +439,33 @@ generate_preprocessed_blocklist_file_parts()
 
 			rm -f /var/run/adblock-lean/blocklist_part_size_B /var/run/adblock-lean/blocklist_part_line_count
 
+			if read -r rogue_element < /var/run/adblock-lean/rogue_element
```
lynxthecat, line 442: But shouldn't we verify that a download completed before trusting the result of the rogue element check (since a partial download could presumably itself result in a rogue element)?

friendly-bits: Should be fixed now, with rogue el. check after download completed check.

lynxthecat:
Can we maybe change the commit message to: 

`Improve error handling and delete obsolete files`

friendly-bits:
> Can we maybe change the title of this pull request and commit message
Done

friendly-bits:
After some thinking, I decided to remove the `cat 1>/dev/null` trick from the following the sed command and instead use the native sed workaround which we already had in place. The reason is unnecessary binary call to cat which will only be useful if rogue elements are detected, which should be pretty rare, and even then performance improvement would be very minimal if any.

adblock-lean
```diff
@@ -490,14 +498,12 @@ generate_preprocessed_blocklist_file_parts()
...
 
-				rm -f /var/run/adblock-lean/rogue_element
-
 				preprocessed_blocklist_line_count=$(( preprocessed_blocklist_line_count + blocklist_file_part_line_count ))
 				blocklist_id=$((blocklist_id+1))
 				continue 2
 			else
+				rm -f "/var/run/adblock-lean/blocklist.${blocklist_id}.gz"
 				log_msg -err "Downloaded blocklist file part line count: $(int2human ${blocklist_file_part_line_count}) less than configured minimum: $(int2human ${min_blocklist_file_part_line_count})."
-				break
```
lynxthecat, line 500: Doesn't this simply avoid retrying the download when we know that the download completed and the downloaded blocklist file part simply doesn't have enough lines in it?

lynxthecat: See: https://forum.openwrt.org/t/optimization-of-adblock-lean/203108/423

PR #50, opened 2024-07-26
Add luci_status function

rickparrish:
This returns status information as a json string, which the luci app can use to display a status table. 
Also modifies two other functions to return unique exit codes for different status conditions.

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -34,6 +34,16 @@ adblock-lean custom commands:
 	gen_config	generate default config
 	update		update adblock-lean to the latest version"
 
+# Result values used by various functions.
+# If values are changed here, they must change in luci-app-adblock-lean as well (overview and status)
+RESULT_DNSMASQ_LOOKUP_FAILED=1
+RESULT_DNSMASQ_LOOKUP_QUADZERO=2
+RESULT_DNSMASQ_NOT_RUNNING=3
+RESULT_NOT_ACTIVE=4
+RESULT_UPDATE_CHECK_ERROR=5
+RESULT_UPDATE_CHECK_LATEST=6
+RESULT_UPDATE_CHECK_OUTDATED=7
+
 mkdir -p /var/run/adblock-lean
 
 trap cleanup_and_exit INT TERM EXIT
@@ -597,7 +607,7 @@ check_dnsmasq()
 	if ! pgrep -x /usr/sbin/dnsmasq &> /dev/null
 	then
 		log_msg -err "No instance of dnsmasq detected with new blocklist."
-		return 1
+		return $RESULT_DNSMASQ_NOT_RUNNING
 	fi
 
 	for domain in google.com amazon.com microsoft.com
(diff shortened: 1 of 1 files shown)


lynxthecat:
Is this ready to merge? Any thoughts @Wizballs or @friendly-bits?

rickparrish:
> Is this ready to merge? Any thoughts @Wizballs or @friendly-bits?

Yes, I believe so.  I submitted as Draft because shell scripting is relatively foreign to me, so while it worked for me wasn't sure if I introduced any other issues!

Although looking at the changes again, I wonder if it would make sense to declare some result code variables and return those instead hardcoding magic numbers.  For example instead of `return 2` to indicate lookup failed, it could do `return $RESULT_LOOKUP_FAILED` or something like that.

friendly-bits:
I'll take a look soon'ish

friendly-bits:
Re. the 3rd patch commit:
I don't mind renaming of the updates check error file. Regarding the other part of the commit which removes the trap, I'd rather not have it in this PR since I'm planning to implement a similar idea which also incorporates a lock file across all functions, and I'd rather have it all in one commit.

Re. the 1st and the 2nd commit: I'll request some changes but it's hard to do when the PR is split into multiple commits which are modifying the same function. I'd suggest @rickparrish to squash those 2 commits into 1 first.

Re. the 4th commit: IMO it's unrelated to this PR. A new PR should be created for it.

friendly-bits:
@rickparrish the way we generally work here is we create separate pull requests (PRs) for separate features/ideas. This helps to work in an organized manner. To prepare a new PR, I create a new "feature" branch in my local fork based on the current master (or from the current development branch when it exists), make commits, then open a pull request from my local branch into master or into the development branch.

rickparrish:
> @rickparrish the way we generally work here is we create separate pull requests (PRs) for separate features/ideas. This helps to work in an organized manner. To prepare a new PR, I create a new "feature" branch in my local fork based on the current master (or from the current development branch when it exists), make commits, then open a pull request from my local branch into master or into the development branch.

Sounds good.  I'm not familiar with squashing, so I'll look into that and then adjust this PR and split the unrelated change into another PR, and will keep what you've said in mind if I have future ideas.

Regarding the lock file across all functions you mentioned, do you expect that would block status/luci_status from running while a blocklist is being downloaded and merged?

friendly-bits:
> Regarding the lock file across all functions you mentioned, do you expect that would block status/luci_status from running while a blocklist is being downloaded and merged?

Ideally status will reflect the current state, so when some operation is ongoing, it will reflect that fact. So in a sense, yes: requesting a status under these conditions will generate something like a "please wait..." response.

friendly-bits:
> I'm not familiar with squashing, so I'll look into that

Check out this answer:
https://stackoverflow.com/a/5201642/23754955

rickparrish:
> > Regarding the lock file across all functions you mentioned, do you expect that would block status/luci_status from running while a blocklist is being downloaded and merged?
> 
> Ideally status will reflect the current state, so when some operation is ongoing, it will reflect that fact. So in a sense, yes: requesting a status under these conditions will generate something like a "please wait..." response.

So then it could differentiate between and report different statuses like Starting, Stopping, Pausing, Resuming, Updating, etc?  That would be perfect.

And I think I jumped the gun on implementing luci_status and the other (quasi-)related changes, so I'm going to close this merge request.  Once your locking/status changes are in place, I'll take a fresh look to see what I really need to provide concise status information on the luci app, and either open a new PR with different changes, or maybe even start with an Issue to discuss my proposed changes and get feedback beforehand.

Also, thanks for the link about squashing!

lynxthecat:
Thanks for the effort here @rickparrish. And also for the constructive feedback @friendly-bits.

PR #52, opened 2024-07-26
Use POSIX syntax for conditionals

friendly-bits:
- Replace `[[`, `]]`, `[[ a == b ]]` with `[`, `]`, `[ a = b ]` to adhere to the POSIX shell scripting standard
- Add shellcheck directives to reduce the shellcheck noise
- (unrelated: specify exit code 1 in a couple of places where we exit because of an error)

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -1,4 +1,5 @@
 #!/bin/sh /etc/rc.common
+# shellcheck disable=SC3043,SC1091,SC3001,SC2018,SC2019,SC3020
 
 # adblock-lean - super simple and lightweight adblocking for OpenWrt
 
@@ -9,7 +10,7 @@
 
 LC_ALL=C
 
-if [[ -t 0 ]]
+if [ -t 0 ]
 then
 	msgs_dest="/dev/tty"
 else
@@ -140,7 +141,7 @@ log_msg()
 log_failure()
 {
 	log_msg -err "${1}"
-	if [[ -n "${report_failure}" ]]
+	if [ -n "${report_failure}" ]
 	then
 		failure_msg="${1}"
 		eval "${report_failure}"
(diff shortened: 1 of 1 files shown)


lynxthecat:
@Wizballs any chance you could give this code a whirl to test?

friendly-bits:
I made an oopsie, line 819 should be:
`if [ ! -f /tmp/dnsmasq.d/.blocklist.gz ] && [ ! -f /tmp/dnsmasq.d/blocklist ]`

I'll force push the fix in a moment

Wizballs:
I test all success and failure conditions, one one tiiiiny little thing ;)
If the Blocklist part is larger than set maximum, "skipping file part and continuing" appears twice:

```
Downloading, checking and sanitizing new blocklist file part from: https://raw.githubusercontent.com/hagezi/dns-blocklists/main/dnsmasq/tif.txt. 
Downloaded blocklist file part size reached the maximum value set in config (15000 KB). 
Consider either increasing this value in the config or removing the corresponding blocklist url. 
Skipping file part and continuing. 
Skipping file part and continuing. 
```

friendly-bits:
We could:
- Either replace in line 451 `break` with `continue 2` - this will skip the download_failed_action check (if we do not consider exceeded filesize a failed download)
- Or remove the message in line 449

Either way, it's unrelated to this PR but I don't mind adding a commit for this.

Wizballs:
Definitely unrelated to POSIX update.   I edited my comment above mentioning that but maybe after you read it.

lynxthecat:
@Wizballs thanks for testing. Is PR OK and ready to merge in?

friendly-bits:
> POSIX update is all good.

@lynxthecat I believe the above part of @Wizballs 's comment answers your question.

Wizballs:
> one config option could generally cover everything (rogue element action, exceeded file size action, failed download action, dnsmasq failed test action).

Actually this makes sense.  Would still no doubt output the relevant reason to the user, but it would reduce the config options.

PR #55, opened 2024-07-29
Add unique return codes to certain functions

rickparrish:
This allows luci-app-adblock-lean to display status information on the overview and app pages.

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -590,6 +590,11 @@ generate_and_process_blocklist_file()
 	return 0
 }
 
+# return values:
+# 0 - dnsmasq is running, and all checks passed
+# 1 - dnsmasq is not running
+# 2 - dnsmasq is running, but one of the test domains failed to resolve
+# 3 - dnsmasq is running, but one of the test domains resolved to 0.0.0.0
 check_dnsmasq()
 {
 	log_msg "Checking dnsmasq instance."
@@ -605,15 +610,15 @@ check_dnsmasq()
 		if ! nslookup "${domain}" 127.0.0.1 &> /dev/null 
 		then
 			log_msg -err "Lookup of '${domain}' failed with new blocklist."
-			return 1
+			return 2
 		elif nslookup "${domain}" 127.0.0.1 | grep -A1 ^Name | grep -q '^Address: *0\.0\.0\.0$'
 		then
 			log_msg -err "Lookup of '${domain}' resulted in 0.0.0.0 with new blocklist."
-			return 1
+			return 3
 		fi
 	done
 	
-	return 0
+	:
 }
 
 restart_dnsmasq()
@@ -812,15 +817,21 @@ gen_stats()
 	log_msg "dnsmasq stats available for reading using 'logread'."
 }
 
+# return values:
+# 0 - adblock-lean is active
+# 1 - adblock-lean is not active
 status()
 {
 	if ! [[ -f /tmp/dnsmasq.d/.blocklist.gz || -f /tmp/dnsmasq.d/blocklist ]]
 	then
 		log_msg "Blocklist in /tmp/dnsmasq.d/ not identified."
 		log_msg "adblock-lean is not active."
-		return
+		return 1
 	fi
-	if check_dnsmasq
+	check_dnsmasq
+	dnsmasq_status=${?}
+	luci_dnsmasq_status=${dnsmasq_status}
+	if [ ${dnsmasq_status} = 0 ]
 	then
 		if [[ -f /tmp/dnsmasq.d/.blocklist.gz ]]
 		then
@@ -829,6 +840,7 @@ status()
 		then
 			good_line_count=$(wc -l /tmp/dnsmasq.d/blocklist)
 		fi
+		luci_good_line_count=${good_line_count}
 		log_msg "The dnsmasq check passed and the presently installed blocklist has good line count: $(int2human ${good_line_count})."
 		log_msg "adblock-lean appears to be active."
 		gen_stats
@@ -837,6 +849,9 @@ status()
 		log_msg "Consider a full reset by running: 'service adblock stop'."
 	fi
 	check_for_updates
+	luci_update_status=${?}
+
+	:
 }
 
 pause()
@@ -855,24 +870,34 @@ resume()
 	log_msg "adblock-lean is now resumed."
 }
 
+# return values:
+# 0 - up-to-date
+# 1 - not up-to-date
+# 2 - update check failed
 check_for_updates()
 {
 	sha256sum_adblock_lean_local=$(sha256sum /etc/init.d/adblock-lean | awk '{print $1}')
 	sha256sum_adblock_lean_remote=$(uclient-fetch https://raw.githubusercontent.com/lynxthecat/adblock-lean/master/adblock-lean -O - 2> /var/run/adblock-lean/uclient-fetch_err | sha256sum | awk '{print $1}')
 
+	local update_check_result
 	if grep -q "Download completed" /var/run/adblock-lean/uclient-fetch_err
 	then
 		if [[ "${sha256sum_adblock_lean_local}" == "${sha256sum_adblock_lean_remote}" ]]
 		then
 			log_msg "The locally installed adblock-lean is the latest version."
+			update_check_result=0
 		else
 			log_msg "The locally installed adblock-lean seems to be outdated."
 			log_msg "Consider running: 'service adblock-lean update' to update it to the latest version."
+			update_check_result=1
 		fi
 	else
 		log_msg -err "Unable to download latest version of adblock-lean to check for any updates."
+		update_check_result=2
 	fi
 	rm -f /var/run/adblock-lean/uclient-fetch_err
+
+	return ${update_check_result}
 }
 
 update()


lynxthecat:
Please can you rename the commit message to: 

`Add unique result codes to certain functions`

to keep it under 50 characters?

lynxthecat:
@Wizballs and @friendly-bits - any feedback for this PR?

friendly-bits:
I already suggested some changes to @rickparrish which he implemented, but now looking at the resulting code, I'd like to suggest some additional changes.

IMO variables used by Luci app should have something like the `luci_` prefix, for example `luci_dnsmasq_status`. This way it's clear from the code why the variable is needed, and will prevent us from accidentally removing or renaming it in the future.

So I'd suggest to change lines 831-834 to this:
```
	luci_dnsmasq_status=1 	# dnsmasq_status is used by luci-app-adblock-lean
	if check_dnsmasq
	then
		luci_dnsmasq_status=0
```

line 850 to this:
`luci_update_status=$?`

Also because shell functions set the return code of the last command as their own return code (unless the `return` command is explicitly issued), it is better to explicitly call the `return` command in function's final line, or have some other way to obviously set the return code. At least for functions where the return code matters. In my own code, I use the `:` command (do-nothing command which always succeeds) in the final line of functions to explicitly set the return code. Otherwise, eventually you end up shuffling some code around and then you accidentally have functions exit with unintended return code. Here, I would suggest the same in the final line of the `status()` and `check_dnsmasq()` functions.

Wizballs:
> variables used by Luci app should have something like the `luci_` prefix
'Best practice' advice such as this is really appreciated again.  Not just here but many times recently.   

A lot of this scripting is above my skill (hobby level), but I can read and learn from all this, so thankyou all.

@lynxthecat 
> any comments for this PR?
This is only a handful of lines to increase luci reporting so I think is good, especially once updated with luci_ naming.

rickparrish:
> Please can you rename the commit message to:
> 
> `Add unique result codes to certain functions`
> 
> to keep it under 50 characters?

Not a problem.  I'll try to keep this in mind for future commits as well.

friendly-bits:
> This would only ever result in luci_dnsmasq_status values of 0 or 1, so would miss out on dnsmasq statuses 2 and 3 (failed to resolve a test domain, and test domain resolved to 0.0.0.0 respectively)

You are right of course - my oversight. So instead perhaps we could have this:
```
	check_dnsmasq
	dnsmasq_status=$?
	luci_dnsmasq_status=${dnsmasq_status}
	if [ $dnsmasq_status = 0 ]
	then
```
I wanted to avoid creating 2 variables but it seems unavoidable.

> Would the inline comment still make sense, or should I drop those since the luci_ prefix will be there?

Perhaps the comment can stay, although it's not critical.


> And after making that change, should I squash the two commits? If I understand what I've read, that will rewrite history, and I wonder if that means the squashed commit will be reflected in this PR, or would this PR essentially become orphaned (referencing a commit that no longer exists) and I would need to open a separate PR?

The way I normally make corrections to a commit which has been submitted in a PR is this. Assuming it's only one commit and that commit is the last one in my local fork, I first modify the code, then issue this command:
`git reset --soft HEAD~1`
This essentially removes the last commit from history while preserving my changes. Then
`git commit <filename> -m <commit_message>`
This creates a new commit with the updated code.
Then
`git push -f`
Basically, a force-push which signals to Github to accept my local commit history which diverged from the remote history. This also automatically updates the commit in the pending PR.

I can't explain this properly because I'm not that good at understanding Git but this works.

rickparrish:
> You are right of course - my oversight. So instead perhaps we could have this:
> 
> ```
> 	check_dnsmasq
> 	dnsmasq_status=$?
> 	luci_dnsmasq_status=${dnsmasq_status}
> 	if [ $dnsmasq_status = 0 ]
> 	then
> ```
> 
> I wanted to avoid creating 2 variables but it seems unavoidable.

Thanks, I've updated my code to match (and also made the other changes you proposed).  

In luci I also make use of $good_line_count, would you suggest also creating a second variable for it as $luci_good_line_count?

> I can't explain this properly because I'm not that good at understanding Git but this works.

That was a good enough explanation for me, thanks!  99% of the time I work solo on projects, so while I use Git daily I don't typically use anything beyond the basic branching, committing, merging aspects.

friendly-bits:
> In luci I also make use of $good_line_count, would you suggest also creating a second variable for it as $luci_good_line_count?

I'd say yes.

The reason is that we have 2 separate code bases here (one is shell code and the other is luci code) and these variables are the only link between them, providing sort of a simple API. So having this API well defined and clear is a good thing.

rickparrish:
> > In luci I also make use of $good_line_count, would you suggest also creating a second variable for it as $luci_good_line_count?
> 
> I'd say yes.
> 
> The reason is that we have 2 separate code bases here (one is shell code and the other is luci code) and these variables are the only link between them, providing sort of a simple API. So having this API well defined and clear is a good thing.

Makes sense to me, so I've added a new line to set `luci_good_line_count`.  I used the same syntax you suggested for setting the dnsmasq status: `luci_good_line_count=${good_line_count}`

On that note, while looking at other scripts to make my changes, I've noticed a few different methods for referencing a variable.  All three of these lines seem to work the same for me, and I'm wondering if deciding between them is personal preference, or are there cases where they would function differently somehow?

```
luci_good_line_count=$good_line_count
luci_good_line_count=${good_line_count}
luci_good_line_count="${good_line_count}"
```

friendly-bits:
> I'm wondering if deciding between them is personal preference, or are there cases where they would function differently somehow?

All 3 lines are functionally equivalent. The 3rd line is better in the sense that quoting variables substitution is a good habit in shell coding, because in some cases values of unquoted variables may be subject to "word splitting" and "globbing". But in this particular case either is not performed (AFAIK at least).

The 2nd line is better in the sense that using the ${var} syntax may be necessary in some cases. E.g. `echo "${var}hello"` is not equivalent to `echo $varhello`, and sometimes if you are not careful and not using the curly braces, you may end up messing something up. I don't practice this because I don't tend to mess up in this particular way. However in adblocklean, this is clearly the convention, so I'm following it here.

Personally, in my code I usually do the 4th variant:
`luci_good_line_count="$good_line_count"`
Because forgetting to quote variables may result in bugs which may be hard to find, so I'd rather have quoting as a habit.

lynxthecat:
Double quoting if in doubt is clearly safer. And presumably most often good practice. 

Albeit in bash (and perhaps also ash) double quoting when unnecessary actually incurs a performance hit because the interpreter escapes every character before interpreting the command. You can see a measure the hit with a simple for loop test. This won’t matter for code like this because there aren’t any tight loops. 

I found this source helpful for knowing when it’s safe not to double quote: 

https://mywiki.wooledge.org/Quotes#When_Should_You_Quote.3F

lynxthecat:
@rickparrish I’m really sorry but would you mind renaming the commit to replace the word result with rename? Sorry I thought I’d specified that but see I hadn’t made that replacement.

rickparrish:
> @rickparrish I’m really sorry but would you mind renaming the commit to replace the word result with rename? Sorry I thought I’d specified that but see I hadn’t made that replacement.

Not a problem, I've renamed it now.  And thanks for the informative link.  I'm going to have to bookmark that, as it's a fair bit to process!

rickparrish:
> Personally, in my own projects I usually do the 4th variant: `luci_good_line_count="$good_line_count"` Because forgetting to quote variables may result in bugs which may be hard to find, so I'd rather have quoting as a habit.

Thank you for taking the time to provide a detailed reply, and the additional variant.  Between this and the link @lynxthecat provided I think I have a much better understanding now.

lynxthecat:
@friendly-bits I think this is OK to merge now?

adblock-lean
```diff
@@ -812,15 +817,21 @@ gen_stats()
...
 	if ! [[ -f /tmp/dnsmasq.d/.blocklist.gz || -f /tmp/dnsmasq.d/blocklist ]]
 	then
 		log_msg "Blocklist in /tmp/dnsmasq.d/ not identified."
 		log_msg "adblock-lean is not active."
-		return
+		return 1
 	fi
-	if check_dnsmasq
+	check_dnsmasq
+	dnsmasq_status=$?
```
lynxthecat, line 832:
```suggestion
	dnsmasq_status=${?}
```

adblock-lean
```diff
@@ -812,15 +817,21 @@ gen_stats()
...
+	luci_dnsmasq_status=${dnsmasq_status}
+	if [ $dnsmasq_status = 0 ]
```
lynxthecat, line 834:
```suggestion
	if [ ${dnsmasq_status} = 0 ]
```

adblock-lean
```diff
@@ -837,6 +849,9 @@ status()
 		log_msg "Consider a full reset by running: 'service adblock stop'."
 	fi
 	check_for_updates
+	luci_update_status=$?
```
lynxthecat, line 852:
```suggestion
	luci_update_status=${?}
```

adblock-lean
```diff
@@ -855,24 +870,34 @@ resume()
...
 			log_msg "Consider running: 'service adblock-lean update' to update it to the latest version."
+			update_check_result=1
 		fi
 	else
 		log_msg -err "Unable to download latest version of adblock-lean to check for any updates."
+		update_check_result=2
 	fi
 	rm -f /var/run/adblock-lean/uclient-fetch_err
+
+	return $update_check_result
```
lynxthecat, line 900:
```suggestion
	return ${update_check_result}
```

lynxthecat:
Oops I missed the lack of curly braces in some places. I think this is the only remaining change before I can merge in.

rickparrish:
> Oops I missed the lack of curly braces in some places. I think this is the only remaining change before I can merge in. Thanks for your patience.

I've just made these updates.  And I should be thanking you for your patience, so thank you!

PR #56, opened 2024-08-02
Implement lock file; improve trap management

friendly-bits:
- Implement lock file check, creation, removal
- Create the trap only for actions which need it
- stop(): kill any running adblock-lean PID's
- status(): check the lock file and if it exists then report on it and exit
- status(): support additional return codes
- Only load config for actions which need it

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -1,5 +1,5 @@
 #!/bin/sh /etc/rc.common
-# shellcheck disable=SC3043,SC1091,SC3001,SC2018,SC2019,SC3020
+# shellcheck disable=SC3043,SC1091,SC3001,SC2018,SC2019,SC3020,SC3003
 
 # adblock-lean - super simple and lightweight adblocking for OpenWrt
 
@@ -35,14 +35,11 @@ adblock-lean custom commands:
 	gen_config	generate default config
 	update		update adblock-lean to the latest version"
 
-mkdir -p /var/run/adblock-lean
-
-trap cleanup_and_exit INT TERM EXIT
-
 cleanup_and_exit()
 {
-	rm -rf /var/run/adblock-lean
 	trap - INT TERM EXIT
+	[ -n "${cleanup_req}" ] && rm -rf /var/run/adblock-lean
+	[ -n "${lock_req}" ] && rm_lock
 	exit
 }
 
@@ -120,6 +117,11 @@ get_elapsed_time_str()
 	printf '%dm:%ds' $((elapsed_time_s/60)) $((elapsed_time_s%60))
 }
 
+print_msg()
+{
+	printf '%s\n' "${1}" > "$msgs_dest"
+}
+
 log_msg()
 {
 	local msg='' _arg='' err_l=info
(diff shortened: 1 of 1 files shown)


adblock-lean
```diff
@@ -915,10 +923,122 @@ update()
...
+		printf '%s\n' "${$} ${action}" > "${pid_file}" || { log_msg -err "Error: Failed to write to pid file '${pid_file}'."; exit 1; }
+	else
+		log_msg -err "Internal error: \${pid_file} variable is unset."
+		exit 1
+	fi
+	:
 }
 
-if [ "${action}" != "help" ] && [ "${action}" != "gen_config" ]
+rm_lock() {
```
lynxthecat, line 950: Would you mind amending the functions to: 

x()
{
…
}

lynxthecat:
Thanks for this. Looks pretty good from a very cursory read through whilst gardening.

adblock-lean
friendly-bits: > Would you mind amending the functions to

Done

friendly-bits:
Changed the function declarations to
```
f ()
{
...
}
```

friendly-bits:
Updated with slightly more compact variable assignment in `report_pid_action()`.

friendly-bits:
^ Yet more tinkering with code compaction in `report_pid_action()` :)
I think that's it.

friendly-bits:
^ Updated with more granular actions set in the lock file.
(and also slightly less spam in the log - we should consider reducing log spam further by utilizing the new `print_msg()` function for some of the less important messages, rather than `log_msg()`)

friendly-bits:
^ Fixed incorrect variable name in `log_msg()` (was `_msg`, should be `msg`).

lynxthecat:
Should be able to test soon. Maybe @Wizballs might also be able to test too.

friendly-bits:
^
- mk_lock(): Only create the directory for the lock file if it doesn't exist yet (otherwise it could have errored out).
- mk_lock(): Remove trailing `.` from the action string when updating the action in the lock file.

Wizballs:
Morning everyone, looks like there has been heaps going on!  Sure thing I'll give everything a test run.

Wizballs:
@lynxthecat Tested multiple different times and service commands with the file lock.  I can't find any failure conditions, so I think this PR is good.

@friendly-bits That's really quite neat how it even outputs what action is currently being performed.  Also I saw another PR where stop/resume/pause was fixed.  Tested these also since I was here, all working well again.

friendly-bits:
^
- Action stop: require cleanup (it wasn't removing `/var/run/adblock-lean/`)
- status(): swap return code 1 and return code 3 for consistency and so return code 1 means 'error' as per standard

(sorry, I found the issue with stop and then I also wanted to change the return codes of status. should be good to go now)

friendly-bits:
^ Change function name from `advert_action` to `reg_action`.

friendly-bits:
^
- Fix restart/reload when adblock-lean is running
- Fix the resume action when blocklist compression is enabled

lynxthecat:
Seems to work well. This evidently required some careful thought and coding. Terrific job @friendly-bits!

PR #58, opened 2024-08-05
Improve error checking and logic

friendly-bits:
(Updated)
- Implement much more error checking
- Implement hierarchical error propagation
-  Move commands initialization code to init_command() and call that function from all main command functions
- Call abl_luci_exit() before exiting (if it's defined by external sourcing script)
- Kill luci.adblock-lean when stopping or restarting
- Get rid of some of the nested logic
- Collect errors in ${failure_msg} and ${luci_errors} and call reg_failure() once before exiting

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -50,17 +50,8 @@ adblock-lean custom commands:
 	gen_config	generate default config
 	update		update adblock-lean to the latest version"
 
-# exit with code ${1}
-# if function 'abl_luci_exit' is defined, execute it before exit
-cleanup_and_exit()
-{
-	trap - INT TERM EXIT
-	[ -n "${cleanup_req}" ] && rm -rf "/var/run/adblock-lean"
-	[ -n "${lock_req}" ] && rm_lock
-	[ -z "${luci_sourced}" ] && [ -n "${report_failure}" ] && [ -n "${failure_msg}" ] && eval "${report_failure}"
-	[ -n "${luci_sourced}" ] && abl_luci_exit "${1}"
-	exit "${1}"
-}
+
+### UTILITY FUNCTIONS
 
 get_file_size_human()
 {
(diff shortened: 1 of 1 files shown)


adblock-lean
```diff
@@ -8,6 +8,12 @@
 # Authors: @Lynx and @Wizballs (OpenWrt forum)
 # Contributors: @antonk; @dave14305 (OpenWrt forum)
 
+# global return codes:
+# 0 - Ok
```
lynxthecat, line 12:
```suggestion
# 0 - OK
```

lynxthecat: Or `Success`?

adblock-lean
```diff
@@ -927,49 +996,55 @@ check_for_updates()
...
 	fi
 	rm -f /var/run/adblock-lean/adblock-lean.latest /var/run/adblock-lean/uclient-fetch_err
 }
 
+# args:
 # 1 - (optional) -f to skip check for existing lock
 # 1/2 - action to write to the pid file
+#
+# return codes:
+# 0 - sucess
```
lynxthecat, line 1018:
```suggestion
# 0 - success
```

friendly-bits:
^ Fixed `sucess`, and `Ok`, Ok? :)

friendly-bits:
^
- manually call status() on 'status' command; call cleanup_and_exit() after status() returns; if sourced then return the correct return code to the sourcing script
(I know that status doesn't need any cleanup but it was more convenient and shorter to call an already existing function, and that function doesn't delete things when not needed)
- log_failure(): register most recent error in `luci_error` variable

friendly-bits:
^
- call `exit` rather than `return` in the code of the main commands
- cleanup_and_exit(): always exit rather than returning
- trap: always exit rather than returning
- move commands initialization code to init_command()
- main command functions: call init_command()
- init_command(): set traps unconditionally
- cleanup_and_exit(): rely on abl_luci_exit() function being defined in sourcing script, rather than on ${abl_sourced} var
- cleanup_and_exit(): call abl_luci_exit() before exit (if defined)

- stop(): accept exit code for argument
- kill_abl_pids(): also kill luci.adblock-lean if it's running

- change var name 'luci_error' -> 'luci_errors'
- log_failure(): collect errors in ${failure_msg} and ${luci_errors} rather than executing report_failure for every error

friendly-bits:
^
- cleanup_and_exit(): check if failure message is set before executing ${report_failure}
- get_stats(): support the '-noexit' argument to return to the calling function
- status(): call gen_stats() with the '-noexit argument'
- rename log_failure() to reg_failure()

adblock-lean
```diff
@@ -140,21 +151,52 @@ log_msg()
...
+		1) reg_failure "Error: failed to export the blocklist."; return 1 ;;
+		2) return 2
+	esac
+	:	
+}
+
+# return codes:
+# 0 - blocklist file present
+# 1 - blocklist file not present
+check_active_blocklist_file()
```
lynxthecat, line 174: Nice sequence!

friendly-bits: You mean the code comments?

friendly-bits:
^
- bytes2human(): fix the comment
- stop(): support optional '-noexit' argument, return to caller when specified
- restart(): implement function (otherwise restart wouldn't work)
- reload(): implement function stub - simply calls restart()
- update_pid_action(): accept rc=3 from check_lock(), reject rc=2
- check_lock(): check if current PID owns the lock file, and if so then return code 3

The above is mainly to fix boot, restart and reload commands

friendly-bits:
^ init_command(): if failed to create/acquire lock, unset ${lock_req}, ${cleanup_req} before exit
(turns out there was a bug where running the start command while another command was running would remove the lock and do cleanup, which should not happen)

lynxthecat:
I think safe to merge this in now?

PR #60, opened 2024-08-12
Improve processing logic

friendly-bits:
- Improve and unify logic of downloaded allowlist and blocklist processing
- Improve console output with highlight colors and spacers
- Implement config validation and parsing
- Change config format
- Implement config migration
- Improve run state checks and reporting
- Remove support for running custom commands specified in config with eval
- Support specifying in config a custom script to source and run functions report_sucess(), report_failure() from
- Support custom dnsmasq directory when specified in UCI
- Append a custom entry to the blocklist and use it to verify that the blocklist was successfully loaded

(I may have forgotten something)

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -463,7 +463,7 @@ cleanup_and_exit()
 	trap - INT TERM EXIT
 	[ -n "${cleanup_req}" ] && rm -rf "${abl_dir}"
 	[ -n "${lock_req}" ] && rm_lock
-	[ -z "${luci_sourced}" ] && [ -n "${failure_msg}" ] && [ -n "${custom_script}" ] && command -v report_failure 1>/dev/null &&
+	[ -z "${luci_sourced}" ] && [ -n "${failure_msg}" ] && [ -n "${custom_scr_sourced}" ] && command -v report_failure 1>/dev/null &&
 		report_failure "${failure_msg}"
 	[ -n "${luci_sourced}" ] && abl_luci_exit "${1}"
 	exit "${1}"
@@ -479,7 +479,7 @@ reg_failure()
 log_success()
 {
 	log_msg "${1}"
-	if [ -n "${custom_script}" ] && command -v report_sucess 1>/dev/null
+	if [ -n "${custom_scr_sourced}" ] && command -v report_sucess 1>/dev/null
 	then
 		report_sucess "${1}"
 	fi
@@ -1165,6 +1165,12 @@ init_command()
 		*)
 			mkdir -p "${abl_dir}" "/tmp/dnsmasq.d"
 			load_config || { reg_failure "Failed to load config."; exit 1; }
+			if [ -n "${custom_script}" ]
+			then
+				custom_scr_sourced=
+				[ -f "${custom_script}" ] && . "${custom_script}" && custom_scr_sourced=1
+				[ -z "${custom_scr_sourced}" ] && report_failure "Custom script '${custom_script}' doesn't exist or it returned error."
+			fi
 	esac
 
 	:
--- a/adblock-lean
+++ b/adblock-lean
@@ -1169,7 +1169,7 @@ init_command()
 			then
 				custom_scr_sourced=
 				[ -f "${custom_script}" ] && . "${custom_script}" && custom_scr_sourced=1
-				[ -z "${custom_scr_sourced}" ] && report_failure "Custom script '${custom_script}' doesn't exist or it returned error."
+				[ -z "${custom_scr_sourced}" ] && reg_failure "Custom script '${custom_script}' doesn't exist or it returned error."
 			fi
 	esac
 
--- a/adblock-lean
+++ b/adblock-lean
@@ -253,6 +253,8 @@ print_def_config()
 # 0 - Success
 # 1 - Error
 # 2 - Unexpected, missing or legacy-formatted (no double quotes) entries found
+#
+# sets the ${missing_keys}, ${needs_fix}, ${luci_conf_needs_fix} variables
 parse_config() {
 	inval_e()
 	{
@@ -265,16 +267,16 @@ parse_config() {
 		needs_fix="${needs_fix}${1}"$'\n'
 	}
 
-	# Following 2 function are needed to minimize ugly hacks and tinkering inside eval
+	# Following 2 functions are needed to minimize ugly hacks and tinkering inside eval
 	# shellcheck disable=SC2317
 	parse_entry()
 	{
 		val=${entry#"${key}="}
 
-		# Following 'case' is a temporarily solution to allow easy config migration - remove a few months from now (Aug 2024)
+		# Following 'case' is a temporary solution to allow easy config migration - remove a few months from now (Aug 2024)
 		case "${entry}" in
 			"compress_blocklist="?" #"*|"initial_dnsmasq_restart="?" #"*)
-				legacy_entries="${legacy_entries}${entry}"$'\n'; legacy_keys="${legacy_keys}${key} "
+				legacy_entries="${legacy_entries}${entry}"$'\n'
 				test_keys="${test_keys%%"${key}|"*}${test_keys#*"${key}|"}"
 				val=${val%% *}
 				return 0
@@ -282,11 +284,17 @@ parse_config() {
 
 		case "${val}" in
 			*\"*\"*\"*) inval_e; return 1 ;; # do not allow more than 2 double-quote marks
-			\"*\"*) ;;
+			\"*\"*)
+				local tmp_val="${val##*\"}" # remove value enclosed in double-quotes
+				case "${tmp_val%%\#*}" in # do not allow characters between 2nd double-quote and in-line comment
+					'') ;;
+					*[!\ ${tab}]*) inval_e; return 1
+				esac
+				;;
 			*\"*\"*) inval_e; return 1 ;; # double quote mark must be the first character
 			*\"*) inval_e; return 1 ;; # do not allow 1 double-quote mark
 			*"#"*) inval_e; return 1 ;; # do not allow in-line comments without double-quote marks
-			*) legacy_entries="${legacy_entries}${entry}"$'\n'; legacy_keys="${legacy_keys}${key} "
+			*) legacy_entries="${legacy_entries}${entry}"$'\n'
 		esac
 		val=${val#\"}
 		val=${val%\"*} # throw away everything after the 2nd double-quote mark
@@ -299,8 +307,8 @@ parse_config() {
 		unexp_entries="${unexp_entries}${entry}"$'\n'
 	}
 
-	local def_config='' conf_format_old='' curr_config='' missing_entries='' unexp_keys='' unexp_entries='' legacy_keys='' legacy_entries='' \
-		curr_config_format def_config_format test_keys entry key val
+	local def_config='' conf_format_old='' curr_config='' missing_entries='' unexp_keys='' unexp_entries='' legacy_entries='' \
+		curr_config_format def_config_format test_keys entry key val tab="$(printf '\t')"
 
 	[ -z "${1}" ] && { reg_failure "parse_config(): no file specified."; return 1; }
 
@@ -391,12 +399,16 @@ parse_config() {
 	fi
 
 	needs_fix="${needs_fix%$'\n'}"
+	luci_conf_needs_fix="${needs_fix}"
 
 	[ -n "${needs_fix}" ] && return 2
 	:
 }
 
-load_config() {
+# shellcheck disable=SC2120
+# 1 - (optional) '-f' to force fixing the config if it has issues
+load_config()
+{
 	local needs_fix='' fixed_config='' missing_keys='' key val line fix cnt \
 		tip_msg="Fix your config file '${config_file}' or generate default config using 'service adblock-lean gen_config'."
 
@@ -409,23 +421,37 @@ load_config() {
 		2)
 	esac
 
-	[ "${msgs_dest}" != "/dev/tty" ] && { log_msg "${tip_msg}"; return 1; }
+	[ "${msgs_dest}" != "/dev/tty" ] && [ "${1}" != '-f' ] && { log_msg "${tip_msg}"; return 1; }
 
 	[ -z "${needs_fix}" ] && { reg_failure "Failed to parse config."; return 1; }
 
-	echo
-	print_msg "${blue}Perform following automatic changes?${n_c}"
-	cnt=0
-	local IFS=$'\n'
-	for fix in ${needs_fix}
-	do
-		[ -z "${fix}" ] && continue
-		cnt=$((cnt+1))
-		print_msg "${cnt}. ${fix}"
-	done
-	IFS="${default_IFS}"
-	pick_opt "y|n" || return 1
-	[ "${REPLY}" = n ] && { log_msg "${tip_msg}"; return 1; }
+	if [ "${msgs_dest}" = "/dev/tty" ]
+	then
+		echo
+		print_msg "${blue}Perform following automatic changes?${n_c}"
+		cnt=0
+		local IFS=$'\n'
+		for fix in ${needs_fix}
+		do
+			[ -z "${fix}" ] && continue
+			cnt=$((cnt+1))
+			print_msg "${cnt}. ${fix}"
+		done
+		IFS="${default_IFS}"
+		pick_opt "y|n" || return 1
+		[ "${REPLY}" = n ] && { log_msg "${tip_msg}"; return 1; }
+	fi
+
+	fix_config "${missing_keys}" || { report_failure "Failed to fix the config."; log_msg "${tip_msg}"; return 1; }
+
+	:
+}
+
+# 1 - missing keys (whitespace-separated)
+fix_config()
+{
+	[ -z "${1}" ] && { reg_failure "fix_config(): invalid argument."; return 1; }
+	local missing_keys="${1}"
 
 	# recreate config from default while replacing values with values from the existing config
 	fixed_config="$(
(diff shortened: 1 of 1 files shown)


Wizballs:
Gave everything a quick test run only and is working great.  I'll run more detailed tests when I can, but this is looking really good.  So many improvements and added functions!

friendly-bits:
^
- parse_config(): detect and prevent previously undetected invalid entries
- parse_config(): support reading and comparing current and default config format versions
- print_def_config(): add config format version
- write_config(): validate new config before moving to permanent storage
- load_config(): don't validate fixed config (now validated by write_config)
- gen_config(): don't validate new config (now validated by write_config)
- load_config(), parse_config(), gen_config(): simplify communication between functions

friendly-bits:
@lynxthecat Would you prefer the commits squashed into one?

lynxthecat:
No please keep them all! But could you maybe capitalise the one beginning “fix.. “ to make it “Fix..” for consistency?

friendly-bits:
^ Fixed the 'fix'

friendly-bits:
^
- parse_config(): compact some code
- parse_config(): don't print legacy keys
- load_config(): save old config before fixing it
- update(): call load_config() upon success

lynxthecat:
@Wizballs I’ll also run a couple of tests and if ours both passes then we can merge in. So far so good for my ordinary day to day runs.

Wizballs:
I'm on now also.  Testing and will report back...

Wizballs:
Working:  
- start/stop/pause/resume 
- rouge check (block and allow)

Not working:
gen_config is just overwriting the config file to defaults.  It's not reusing existing valid config entries


And two items that are for later maybe (one discussed previously):

If a cron job is called while an invalid config exists, it correctly exists.  Should it report_failure so people can receive a message? 

```
user.info Fix your config file '/root/adblock-lean/config' or generate default config using 'service adblock-lean gen_config'.
user.err adblock-lean: Error: Failed to load config.

```

Discussed previously, but will the custom script have a default path, and have a config entry disable by default?  This would save editing the path on a gen_config, could just switch to a YES/1 or whatever the config entry would be.

friendly-bits:
> * gen_config is just overwriting the config file to defaults.  It's not reusing existing valid config entries

"It's not a bug, it's a feature". In what scenario would you want gen_config() to reuse existing values? My logic is this: if user's config is technically compliant to the format then there is no need to fix it. If it's not compliant then load_config() will pick this up and suggest fixes. If the user is unhappy with their config for any reason, or if the config file is missing, or if the config file has issues which can not be automatically fixed, then they can discard the old config and have it replaced with the default one - this is what gen_config() is for.

Wizballs:
> , it's a feature

Perfect got it.  I'll check the load_config part....

Wizballs:
Alright I got it now.  Intentionally deleted lines and messed up my config.  And it recommended the missing lines and highlighted entries that shouldn't be there.  Worked great on multiple different test.

Very smooth.

friendly-bits:
> * will the custom script have a default path, and have a config entry disable by default?  This would save editing the path on a gen_config, could just switch to a YES/1 or whatever the config entry would be.

This makes sense. Pro's: lower chance of user error when copy-pasting the path, slightly easier process for the user. Con's: adds another config option. I don't mind either way.

friendly-bits:
> * If a cron job is called while an invalid config exists, it correctly exists.  Should it report_failure so people can receive a message?

Checking on this... Will comment a bit later.

Wizballs:
I think, leave this as is.  Keep config options lower.  Anyone using a custom script should be able to manage a simple 
/root/adblock-lean/custom_script as per Ree9's Luci default.

Sometimes I just have to talk through things to get it right sorry.  But in that case, should there be a comment mentioning this default path for use later on with Luci?

friendly-bits:
> * If a cron job is called while an invalid config exists, it correctly exists.  Should it report_failure so people can receive a message?

So the challenge here is this: if the config is invalid, can we trust it enough to source the custom script and call the report_failure() function from it? Current approach relies on valid config as 'root-of-trust' for the matter of safely using any config entries. Technically we could implement an exclusion for the custom_script option but I'm not sure if this is a good idea. Not saying that it's not. Just not sure.

lynxthecat:
> "It's not a bug, it's a feature". In what scenario would you want gen_config() to reuse existing values? My logic is this: if user's config is technically compliant to the format then there is no need to fix it. If it's not compliant then load_config() will pick this up and suggest fixes. If the user is unhappy with their config for any reason, or if the config file is missing, or if the config file has issues which can not be automatically fixed, then they can discard the old config and have it replaced with the default one - this is what gen_config() is for.

Just so I understand, valid old entries will be retained, but invalid old entries will be deleted and any needed replacements adds as defaults, right? So a custom config element without double quotes will get lost, but a custom config element with double quotes will be retained? I think that's probably just fine.

Wizballs:
> can we trust it enough to source the custom script and call the report_failure()

Looking at this another way also, if a custom notification does come through, that should be reason alone to investigate what is happening.  Leave as, I actually don't think it's over important or worth additional lines of code.   Thanks for bouncing idea's back and forth, sometimes it just helps to put a reason behind it.

lynxthecat:
Seems super daring and ambitious to have config elements automatically changed, but it does seem to be working. Nice one.

Wizballs:
> config elements automatically changed, but it does seem to be working. Nice one

It works really well.  I messed my config up various different ways, and it always fixed missing, or reported "rogue" entries ;)

friendly-bits:
While we're at it, maybe now it's a good time to stop using /root and move stuff to where it should normally be? I.e. probably to /etc/adblock-lean/.

friendly-bits:
> Seems super daring and ambitious to have config elements automatically changed

It looks more impressive than it actually is. The logic is very simple. We check for critical issues which would deem the config unusable - if found, we throw an error and quit. If not found then we load the values for expected keys from the existing config, while registering missing and unexpected keys, or entries which don't have double quotes around the values. Then we report of any found issues to the user and ask their permission to fix them. The fix is ultra-simple: we take the default config file and replace the values in it with values of the known-good entries in the current config. Then we validate the result. If it passes validation, we write it out. That's it.

Wizballs:
Well I've thrown everything at this, and it's come up spot on.  Handles every situation / config as it should.   Awesome!

To get the success or failure notification, I had to add the export success_msg="${1}" &&
Is this the right way or is there a better way?

log_success()
```
{
	log_msg "${1}"
	if [ -n "${custom_scr_sourced}" ] && command -v report_success 1>/dev/null
	then
		export success_msg="${1}" && report_success "${1}"
	fi
}
```

friendly-bits:
The only shortcoming of this method (which I'm aware of) is that custom comments are not preserved after fixing the config. You essentially get a vanilla config with your existing values for valid keys. Moving the custom comments would be really challenging, because the comments are probably attached to entries in some way which a simple algorithm has no means to figure out.

friendly-bits:
> Just so I understand, valid old entries will be retained, but invalid old entries will be deleted and any needed replacements adds as defaults, right? So a custom config element without double quotes will get lost, but a custom config element with double quotes will be retained? I think that's probably just fine.

Oops, just noticed this comment.

> valid old entries will be retained
Yes
> old entries will be deleted
Yes
> any needed replacements adds as defaults, right?
The fix-related code essentially replaces default values with existing values extracted from valid entries in the existing config (including values extracted from entries in "legacy format", i.e. missing double-quotes). Missing entries entries are copied over from the default config.
> a custom config element without double quotes will get lost
An entry which has its value not enclosed in double-quotes is considered "legacy format". The value is still retained when the config is fixed but the config must be fixed before it can be used.

An exception to this are entries which both don't have values enclosed in double quotes, and the value contains the `#` character, because in this situation we don't have a reliable way to determine whether `#` is a beginning of a comment or a part of a string (i.e. URL). 

An exception to the exception (I know, it's a bit confusing) is a special case of 2 entries which exist in our older default config, don't have double-quotes and have in-line comments. Specifically:
```
compress_blocklist=1 # enable (1) or disable (0) blocklist compression
initial_dnsmasq_restart=0 # enable (1) or disable (0) initial dnsmasq restart
```
The following code in parse_entry() specifically addresses these entries so they are not flagged as invalid but rather as legacy:
```
		# Following 'case' is a temporarily solution to allow easy config migration - remove a few months from now (Aug 2024)
		case "${entry}" in
			"compress_blocklist="?" #"*|"initial_dnsmasq_restart="?" #"*)
				legacy_entries="${legacy_entries}${entry}"$'\n'; legacy_keys="${legacy_keys}${key} "
				test_keys="${test_keys%%"${key}|"*}${test_keys#*"${key}|"}"
				val=${val%% *}
				return 0
		esac
```

friendly-bits:
^
- parse_config(): set ${luci_conf_needs_fix}
- parse_config(): fix a couple typos
- parse_config(): remove ${legacy_keys) (unused variable)
- parse_config(): catch another previously uncaught case of invalid entries
- fix_config(): move config-fixing code to a separate function
- load_config(): accept the '-f' argument to force fixing config (bypassing console dialog and /dev/tty check)
- load_config(): call fix_config()

The main change here is exposing the required fixes to luci via ${luci_conf_needs_fix} variable, and added functionality of forced config fix via a call to `load_config()` with the argument `-f`. Both are intended to allow the luci app to have a similar dialog with the user as currently implemented via console, and initiate the fixes. The idea is something like:
1) call parse_config() (needs path to the conf_file)
2) check return code - if it's 2 then fixes are needed
3) pull suggested fixes from the ${luci_conf_needs_fix} variable
4) to initiate config fix, call `load_config -f`

Not sure if @rickparrish would want to use it but I thought it could be a good idea to make this functionality available.

rickparrish:
> Not sure if @rickparrish would want to use this but I thought it could be a good idea to make the functionality available.

Sounds like an excellent idea, I'll definitely make use of it.

The luci app does have logic for updating a config file while keeping settings (eg remapping `min_blocklist_file_part_line_count` to `min_blocklist_part_line_count`), but now that you have the logic in adblock-lean itself it's redundant to have it in the luci app.  So removing it will streamline the code a bit, and minimize the amount of testing needed for future config changes.

rickparrish:
I now have a `Reset` button appearing when `parse_status` returns 1 that will call `gen_config`, and an `Update` button appearing when `parse_status` returns 2 that will call `load_config "-f"`

I tried several different config files, and most worked, but I did run into an issue with one.  You should be able to replicate it by doing:
- `service adblock-lean gen_config`
- Edit the config file to remove the `# config_format=v2` line
- `service adblock-lean start`

This is the output I get:
```
root@OpenWrt:~/adblock-lean# service adblock-lean start

Warning: Config format version is unknown or invalid.

Perform following automatic changes?
1. Update config format version
y|n: y
Error: fix_config(): invalid argument.
/etc/rc.common: line 445: report_failure: not found
Fix your config file '/root/adblock-lean/config' or generate default config using 'service adblock-lean gen_config'.
Error: Failed to load config.
```

friendly-bits:
> I did run into an issue with one

Good catch. The check for ${missing_keys} should not have been there. Removed with the above commit ^.

friendly-bits:
^ exposes additional variables to the luci RPC script: ${luci_unexp_keys}, ${luci_unexp_entries}, ${luci_missing_keys}, ${luci_missing_entries}, ${luci_legacy_entries}, ${luci_bad_conf_format}

This in addition to previously exposed variables:
${luci_curr_config_format}, ${luci_def_config_format},  ${luci_conf_needs_fix}

Is this useful to you @rickparrish ?

friendly-bits:
^
- check_active_blocklist(): perform separate check for blocklist test domain and assign separate return value if it fails
- status(): suggest running 'restart' (rather than 'stop') if full reset is required
- status(): exit with code 1 if check for active blocklist fails

rickparrish:
> Is this useful to you @rickparrish ?

Yeah that sounds perfect, that'll allow me to relay to the user what changes they need to make if they want to do it manually.  I meant to display ${luci_conf_needs_fix} but forgot all about it, but now I'm glad I didn't because I can jump straight into using the individual variables instead.

friendly-bits:
> I can jump straight into using the individual variables instead

I think both may be useful. The newly-exposed variables say what issues were detected, while the ${luci_conf_needs_fix} variable lists changes that will be performed by the automatic fix. To you discretion of course.

friendly-bits:
^ Since the number of luci_ variables is growing, I think it no longer makes sense to list them all in the comment at the beginning of  the script. I hope it's OK with you @rickparrish.

friendly-bits:
I think at this point the PR is ready.

rickparrish:
> I think both may be useful. The newly exposed variables say which issues were detected, while the ${luci_conf_needs_fix} variable lists changes that will be performed by the automatic fix. To you discretion of course.

Ahh, I thought the single variable was essentially a combination of the other separate variables, but yes it sounds like all of them will be useful.  Just doing some additional testing of package building, and then will get the "update needed" page working with these variables.

> ^ Since the number of luci_ variables is growing, I think it no longer makes sense to list them all in the comment at the beginning of the script. I hope it's OK with you @rickparrish.

Yes, that's fine with me.  I've been keep tracking of them in the RPC script as well, so I'll still have a quick reference of what's available, and what the different values mean.

friendly-bits:
> and what the different values mean

Also note that `check_active_blocklist()` now has a separate return value for when the blocklist test domain fails to resolve (which means essentially that the blocklist was not loaded). If you are using that function, of course.

friendly-bits:
^ Renamed ${needs_fix} to ${conf_fixes}, ${luci_conf_needs_fix} -> ${luci_conf_fixes}, for clarity.

rickparrish:
> ^ Renamed ${needs_fix} to ${conf_fixes}, ${luci_conf_needs_fix} -> ${luci_conf_fixes}, for clarity. @rickparrish

Thanks, got this updated on the Config Update Needed screen now.  

And yeah, now that it's in place there was definitely value in both ${luci_conf_fixes} and the other new variables you added, since it can inform the user what will happen if they click the button, or what they need to do if they want to make the changes manually.

friendly-bits:
^
- rm_allow_domains(): add function
- rm_allow_domains(): fix allowlist domains removal from blocklist
- download_and_process_list_part(): use rm_allow_domains()
- gen_list_parts(): use rm_allow_domains() on local blocklist
- generate_and_process_blocklist_file(): fix lines count
- status(): fix lines count

This fixes some bugs:
- allow domains not removed from downloaded blocklist
- allow domains not removed from local blocklist
- lines count was still off in generate_and_process_blocklist_file()
- lines count was off in status()

Wizballs:
The removal of allowlist from local blocklist makes sense, that wasn't happening previously.  Removal of allowlist was already working though, so just out of curiosity, what scenario was fixed or refined for that?

Wizballs:
Ah yes I see allowlist now has the server=/[domain]/# applied after download now, instead of piped in before the sort -u command.   Looking forward to trying this one, will have to wait until first thing in the morning now.  Shall let you know!

adblock-lean
```diff
@@ -129,60 +137,418 @@ print_msg()
...
 	[ -z "${1}" ] || [ -z "${2}" ] && { reg_failure "try_mv(): bad arguments."; return 1; }
-	mv -f "${1}" "${2}" || { reg_failure "Error: failed to rename '${1}'."; return 1; }
+	mv -f "${1}" "${2}" || { reg_failure "Failed to rename '${1}'."; return 1; }
 	:
 }
 
+# asks the user to pick an option
+# 1 - input in the format 'a|b|c'
+# output via $REPLY
+pick_opt() {
```
lynxthecat, line 167: Would it be OK to put the { on a new line?

adblock-lean
```diff
@@ -129,60 +137,418 @@ print_msg()
...
+#
+# return codes:
+# 0 - Success
+# 1 - Error
+# 2 - Unexpected, missing or legacy-formatted (no double quotes) entries found
+#
+# sets ${missing_keys}, ${conf_fixes}
+# and variables for luci:
+# *_curr_config_format *_def_config_format *_unexp_keys *_unexp_entries *_missing_keys *_missing_entries *_legacy_entries *_bad_conf_format *_conf_fixes
+parse_config() {
```
lynxthecat, line 255: Can we put the `{` onto a new line?

adblock-lean
```diff
@@ -33,6 +31,9 @@ else
 fi
 
 PREFIX=/root/adblock-lean
+abl_dir=/var/run/adblock-lean
+tmp_abl_dir=/tmp/adblock-lean
```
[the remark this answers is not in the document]
friendly-bits: Initially the idea was to use this directory for the lock file. Storing the lock file in `/var/run/adblock-lean` intuitively seems wrong because (for example) if the lock file was stored in the same directory and we had 2 instances of adblock-lean running (for example one is downloading and applying the blocklist, another one is called with the status command), a subtle mistake in the code could lead to race conditions, for example one instance could delete the directory and the lock file along with it while the other instance expects the lock file to be present. It is possible to keep this in mind when coding and carefully test for such conditions, but much easier to just keep a separate directory for the lock file.

Apparently later I got confused and used that directory in the write_config() function. This is a mistake, I'll fix this in a moment.

friendly-bits:
^
- rename ${tmp_abl_dir} -> ${abl_pid_dir} to avoid confusion
- write_config(): use ${abl_dir} for temporary config file
- improve code style (`while ... do`, `pick_opt() {`)

Wizballs:
@friendly-bits FYI all allowlist/blocklist testing completed with the new code flow, and everything working as expected.

friendly-bits:
^
- remove fix_line_cnt()
- download_and_process_list_part(), gen_list_parts(), generate_and_process_blocklist_file(), status(): don't add newlines when processing lists; use 'wc -w'

PR #65, opened 2024-08-18
Consolidate processing of local and downloaded list parts

friendly-bits:
- download_and_process_list_part() - rename function to process_list_part()
- rm_allow_domains(): remove function - code integrated into process_list_part()
- process_list_part(): add functionality to process local list parts
- process_list_part(): limit logging dnsmasq errors to 10 first lines
- gen_list_parts(): log_process_sucess(): add function
- gen_list_parts(): handle_process_failure(): add function
- gen_list_parts(): adapt to call process_list_part() for both local and downloaded lists
- gen_list_parts(): delete downloaded allowlist parts after merging them into the main allowlist file
- init_command(): reload, restart actions: don't set cleanup_req=1 kill_req=1 lock_req=1 (requirements are handled when initializing stop() and start() functions)
- init_command(): reload, restart actions: don't detect dnsmasq dir (detection is handled when initializing stop() and start() functions)
- init_command(): reload, restart actions: don't call check_dnsmasq_instance (check is handled when initializing the start() function and shouldn't be performed before calling the stop() function)
- some improvements to console output (additional spacers, more laconic messages)

Main changes are in gen_list_parts(), process_list_part(). The init_command() changes are intended to a. avoid performing unnecessary operations when action is reload or restart, and b. to avoid calling check_dnsmasq_instance() when action is restart or reload, because otherwise that check may fail if dnsmasq crashed because of a prior issue, which would cause the restart or reload action to fail, which would be unwanted.

Needs more testing, of course.

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -449,7 +449,7 @@ load_config()
 
 	[ -z "${conf_fixes}" ] && { reg_failure "Failed to parse config."; return 1; }
 
-	if [ "${msgs_dest}" = "/dev/tty" ]
+	if [ "${msgs_dest}" = "/dev/tty" ] && [ "${1}" != '-f' ]
 	then
 		echo
 		print_msg "${blue}Perform following automatic changes?${n_c}"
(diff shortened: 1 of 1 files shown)


lynxthecat:
I haven't been able to break it yet. Maybe @Wizballs will have more success.

friendly-bits:
^
- load_config(): correctly skip dialog when called with the '-f' argument
- update(): support the '-no-check-config' argument to skip the config compatibility check and config conversion (intended to give control over the process to the luci app)
- generate_and_process_blocklist_file(): don't remove empty lines

friendly-bits:
^
- process_list_part(): limit size of local lists as well
- generate_and_process_blocklist_file(): improve console message
- generate_and_process_blocklist_file(): count bytes and words in one go
- generate_and_process_blocklist_file(): remove useless cats

friendly-bits:
^
- update(): log_fallback(): rename to failsafe_log()
- update(): failsafe_log(): simplify implementation

friendly-bits:
Yes, as far as I know :)

PR #67, opened 2024-08-20
Fix stop action when sleeping after boot

friendly-bits:
- kill_abl_pids(): detect and kill adblock-lean process when run from the `/etc/rc.d` path
- kill_abl_pids(): kill child processes spawned by the main adblock-lean process (eg `sleep`)
- init_command(): unset ${cleanup_req}, ${lock_req} before exit if failed to kill adblock-lean processes (fixes erroneous removal of the lock file and the /var/run/adblock-lean directory)

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -1192,16 +1192,20 @@ check_lock()
 # kills any running adblock-lean instances
 kill_abl_pids()
 {
-	local _killed _p _pid IFS=$'\n' k_attempt=0
+	local _killed _p _pid child_pid IFS=$'\n' k_attempt=0
 	while :
 	do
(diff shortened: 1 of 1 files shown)


friendly-bits:
This is a pretty simple commit, not much testing is required.

lynxthecat:
How would you recommend we test this one?

friendly-bits:
To the extent of what I can think of, I have already tested it. If you want to test same things then use the `stop` action under different conditions, including while adblock-lean is sleeping after boot (which was failing before this commit), and see if you get any errors or obvious issues.

lynxthecat:
Ah. Sounds good. I'll just merge. Livin' life on the edge this week.

PR #133, opened 2025-03-29
Parallel processing and other improvements

friendly-bits:
This PR contains all the accumulated commits in the `simpler_concurrency` branch (after I managed to rebase it on the current master). Short version, in no particular order:

- Implement parallel downloads and processing
- Add option `MAX_PARALLEL_JOBS`, default to `auto`
- Replace option `initial_dnsmasq_restart` with `unload_blocklist_before_update`, default to `auto`
- parse_config(): rewrite parser in awk
- abl_process: skip dnsmasq restart if adblock-lean run state is stopped
- abl_process: process_list_part(): use one `wc` call
- Multiple commits improving error detection and handling across scripts
- fix `status` command reported entries count being off by 1
- Use a separate file for registering and checking current action (previously the PID file was used for that)
- Write to log file asynchronously
- Multiple changes to console output messages across scripts
- log_msg(): output empty lines to $MSGS_DEST (fixes empty log entries spam at boot)
- check_lock(): retry reading the PID file if read fails
- init_command: require lock for action boot (fixes `status` command not reporting that adblock-lean is sleeping after boot)
- abl_post_update_1: set actually correct stop priority
- Make multiple attempts when resolving domains
- abl-process: cope with filesystem delays
- abl-install: offer setup if conf file not present
- Avoid subshell when calling int2human
- Print uclient-fetch errors where relevant
- explicitly call busybox applets gzip, gunzip, zcat

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -279,6 +279,7 @@ cleanup_and_exit()
 	trap - INT TERM EXIT
 	if [ -n "${CLEANUP_REQ}" ]
 	then
+		[ -n "${WATCHDOG_PID}" ] && kill -9 "${WATCHDOG_PID}" 2>/dev/null
 		[ -n "${SCHEDULER_PID}" ] && [ ! -f "${SCHEDULE_DIR}/scheduler_done_${SCHEDULER_PID}" ] &&
 		{
 			log_msg -warn "Killing unfinished processing jobs."
--- a/adblock-lean
+++ b/adblock-lean
@@ -608,11 +608,13 @@ kill_pids_recursive()
 
 # return codes:
 # 0 - running
-# 1,2 - (reserved)
+# 1 - error
+# 2 - (reserved)
 # 3 - paused
 # 4 - stopped
 get_abl_run_state()
 {
+	[ -n "${DNSMASQ_CONF_D}" ] || { reg_failure "\$DNSMASQ_CONF_D is not set"; return 1; }
 	local f
 	for f in "${DNSMASQ_CONF_D}/.abl-blocklist.gz" "${DNSMASQ_CONF_D}/abl-blocklist"
 	do
@@ -1241,6 +1243,7 @@ status()
 	log_msg -purple "" "adblock-lean (${version}) status:"
 	case ${run_state} in
 		0) ;;
+		1) reg_failure "Failed to check adblock-lean run state." ;;
 		3) log_msg -yellow "" "adblock-lean is paused." ;;
 		4) log_msg -yellow "" "adblock-lean is stopped."
 	esac
@@ -1263,7 +1266,7 @@ status()
 			luci_good_line_count=${active_entries_cnt}
 			int2human active_entries_cnt_human "${active_entries_cnt}"
 			log_msg "The dnsmasq check passed and the presently installed blocklist has entries count: ${active_entries_cnt_human}."
-			log_msg -green "adblock-lean is active."
+			log_msg -green "" "adblock-lean is active."
 			gen_stats -noexit
 		else
 			reg_failure "The dnsmasq check failed with existing blocklist file."
@@ -1300,6 +1303,7 @@ resume()
 	get_abl_run_state
 	case ${?} in
 		0) log_msg -err "adblock-lean is already running."; exit 1 ;;
+		1) exit 1 ;;
 		3) ;;
 		4) log_msg -err "adblock-lean is currently stopped, not paused. Can not resume."; exit 1;
 	esac
--- a/adblock-lean
+++ b/adblock-lean
@@ -442,6 +442,7 @@ update_action_file() {
 	[ -z "${1%.}" ] && { reg_failure "${me}: action is unspecified."; return 1; }
 
 	{
+		try_mkdir -p "${ABL_PID_DIR}" &&
 		printf '%s\n' "${1%.}" > "${ACTION_FILE}" || reg_failure "${me}: Failed to write to action file '${ACTION_FILE}'."
 	} & # runs asynchronously
 	:
--- a/adblock-lean
+++ b/adblock-lean
@@ -139,6 +139,55 @@ add2list() {
 	:
 }
 
+# reads first line from file
+# Args:
+# -v <out_var_name>
+# -f <file_path>
+# Optional: '-a <attempts_num>'
+# Optional: '-d' to delete the file after reading
+# Optional: '-q' to quiet error message
+# Optional: '-n X' to limit number of bytes read
+# Optional: '-D <"[file_desc]">'
+# Optional: '-V <"[default_val]">'
+read_str_from_file()
+{
+	local me=read_str_from_file rs_del_file='' rs_quiet='' rs_num_bytes='' rs_outvar='' rs_file='' \
+		rs_file_desc='' rs_file_desc_pr='' rs_def_val='' rs_read_failed='' rs_attempts=1 rs_attempt
+	while getopts ":dqn:v:f:a:D:V:" opt
+	do
+		case "${opt}" in
+			d) rs_del_file=1 ;;
+			q) rs_quiet=1 ;;
+			n) rs_num_bytes=-n${OPTARG} ;;
+			v) rs_outvar=${OPTARG} ;;
+			f) rs_file=${OPTARG} ;;
+			a) rs_attempts=${OPTARG} ;;
+			D) rs_file_desc=${OPTARG} ;;
+			V) rs_def_val=${OPTARG} ;;
+			*) reg_failure "${me}: unexpected option '${opt}'."; return 1;
+		esac
+	done
+
+	[ -n "${rs_outvar}" ] && [ -n "${rs_file}" ] || { reg_failure "${me}: missing args"; return 1; }
+
+	rs_attempt=0
+	while :
+	do
+		[ -f "${rs_file}" ] && IFS="" read -r ${rs_num_bytes?} "${rs_outvar?}" < "${rs_file}" && eval "[ -n \"\${${rs_outvar}}\" ]" && break
+		rs_attempt=$((rs_attempt+1))
+		[ "${rs_attempt}" -le "${rs_attempts}" ] || { rs_read_failed=1; break; }
+		sleep 1
+	done
+
+	[ -n "${rs_del_file}" ] && rm -f "${rs_file}"
+	[ -z "${rs_read_failed}" ] && return 0
+
+	[ -n "${rs_file_desc}" ] && rs_file_desc_pr="${rs_file_desc} "
+	[ -n "${rs_quiet}" ] || reg_failure "Failed to read ${rs_file_desc_pr}file '${rs_file}'."
+	eval "${rs_outvar}=\"${rs_def_val}\""
+	return 1
+}
+
 get_md5()
 {
 	md5sum "${1}" | cut -d' ' -f1
@@ -364,20 +413,7 @@ check_lock()
 	[ -z "${PID_FILE}" ] && { reg_failure "\${PID_FILE} variable is unset."; return 1; }
 	[ ! -f "${PID_FILE}" ] && return 0
 
-	local read_failed='' read_attempt=0
-	while :
-	do
-		IFS="" read -r LOCK_PID < "${PID_FILE}" && break
-		read_attempt=$((read_attempt+1))
-		[ "${read_attempt}" -le 3 ] || { read_failed=1; break; }
-		sleep 1
-	done
-
-	[ -n "${read_failed}" ] &&
-	{
-		reg_failure "Failed to read the pid file '${PID_FILE}'."
-		return 1
-	}
+	read_str_from_file -v LOCK_PID -f "${PID_FILE}" -a 3 -D PID || return 1
 
 	case "${LOCK_PID}" in
 		"${$}") return 3 ;;
@@ -429,8 +465,7 @@ report_curr_action()
 	fi
 	reported_pid="${LOCK_PID:-unknown}"
 
-	IFS="" read -r PID_ABL_CMD < "${ACTION_FILE}"
-	: "${PID_ABL_CMD:="unknown action"}"
+	read_str_from_file -v PID_ABL_CMD -f "${ACTION_FILE}" -a 2 -D action -V "unknown action"
 
 	[ "${1}" = -log ] && report_cmd=log_msg
 
@@ -936,11 +971,7 @@ install_abl_files()
 	local dist_dir="${1}" version="${2}"
 
 	# read new files list
-	read -r new_files < "${dist_dir}/inst_files" ||
-	{
-		reg_failure "Failed to read file '${dist_dir}/inst_files'."
-		return 1
-	}
+	read_str_from_file -d -v new_files -f "${dist_dir}/inst_files" -a 2 || return 1
 
 	# compile current files list
 	curr_files="${ABL_SERVICE_PATH}"
--- a/adblock-lean
+++ b/adblock-lean
@@ -139,7 +139,7 @@ add2list() {
 	:
 }
 
-# reads first line from file
+# reads first line from file into variables
 # Args:
 # -v <out_var_name>
 # -f <file_path>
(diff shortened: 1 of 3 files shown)


adblock-lean
```diff
@@ -454,35 +533,88 @@ log_success()
...
+		for pid in ${running_pids}
+		do
+			[ -d "/proc/${pid}" ] && alive_pids="${alive_pids}${pid} "
+		done
+		running_pids="${alive_pids}"
+		[ -n "${running_pids}" ] || return 0
+
 		sleep 1
 	done
+	kill -9 "${running_pids}" 2>/dev/null
```
lynxthecat, line 605: Will this only run providing the earlier graceful kills did not work? I'm thinking about recyling of PID values to different processes. Or at least killing by fire should only be last resort.

friendly-bits: The implementation makes 1 initial attempt at normal kill and then 10 more attempts, with 1 sec pause after each. Before retrying, it checks for PIDs which survived. Only those PIDs are sent the kill signal at the next attempt. If some survived after 10 attempts, they are sent the kill -9 signal. As far as I can tell, the kernel does not recycle PIDs.

usr/lib/adblock-lean/abl-process.sh
```diff
@@ -1,254 +1,231 @@
...
+
+ABL_TEST_DOMAIN="adblocklean-test123.info"
+
+
+# UTILITY FUNCTIONS
 
 try_gzip()
 {
-	gzip -f "${1}" || { rm -f "${1}.gz"; reg_failure "Failed to compress '${1}'."; return 1; }
+	busybox gzip -f "${1}" || { rm -f "${1}.gz"; reg_failure "Failed to compress '${1}'."; return 1; }
```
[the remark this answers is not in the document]
friendly-bits: No, but this is guaranteed to work. I had issues where I would uninstall gnu utilities and then the system couldn't find the busybox alternatives unless called explicitly.

adblock-lean
friendly-bits: Also this function is only called by stop(). If the jobs or the scheduler need to be stopped, this is handled separately (just sending the normal kill signal to relevant PIDs).

usr/lib/adblock-lean/abl-process.sh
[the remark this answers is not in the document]
friendly-bits: Admittedly, the issues were in an (already kinda outdated) Openwrt snapshot which I'm using for testing, so maybe this is fixed in the released version, but in general I think this is the better practice (and maybe we should even explicitly provide the path to the busybox binary).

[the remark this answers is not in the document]
friendly-bits: In terms of performance, I compared busybox gzip vs gnu gzip and didn't see any difference. So IMO even if another implementation of gzip is installed, we can probably ignore it. The only potential issue I can see is customized OpenWrt build which disables the Busybox gzip applets. Probably nobody is using such build but if someone does and they report an issue, we can always implement detection and replace the calls with something like `${GZIP_CMD}`.

friendly-bits:
In the meantime, I discovered some additional minor issues with this PR:
- current action will only be read when `${MSGS_DEST}` is not `/dev/null`, which will prevent the luci app from knowing which action is being currently preformed.
- The config parser, while validating the default config, ignores the value set in default config - this is wrong and can be easily fixed, but will only matter when we make a mistake in the default config.
- possibly some additional improvements can be made to the config parser
- I'm not entirely sure that the function `kill_pids_recursive()` is really needed. It implements child PIDs discovery and this makes it somewhat slow. While it is only called in the context of the `stop()` command, `stop()` is called as part of quite a few other actions and possibly a simpler and/or smarter way to handle this can be implemented.

None of this is critical and I think this should not block merging to master (I'll remind that the current update mechanism doesn't automatically fetch updates from master anymore). So I think I'll hold on with addressing these till this PR is merged.

adblock-lean
```diff
@@ -454,35 +533,88 @@ log_success()
...
+			add_child_pids running_pids "${pid}"
 		done
-		[ -z "${_killed}" ] || [ ${k_attempt} -gt 10 ] && break
+
+		kill "${running_pids}" 2>/dev/null
+
+		local alive_pids=''
+		for pid in ${running_pids}
+		do
+			[ -d "/proc/${pid}" ] && alive_pids="${alive_pids}${pid} "
```
lynxthecat, line 598: Isn't this a case for `kill -0`? 

See: https://unix.stackexchange.com/a/169899

friendly-bits: This is another way of implementing this but do we have a reason to believe that this is faster?

friendly-bits:
> Could you be convinced?

I could, actually was considering something similar. My issue though is that the current method is basically bullet-proof. It will kill the process and any child processes, no matter what. While sending a signal to a process doesn't guarantee that the process actually exits. It will exit under normal circumstances but not, for example, while it's sleeping (it does actually receive the signals immediately when it's executing the `wait` command, from my testing). So then we will need implement all sufficiently long sleep commands like so:
```
sleep [X] & # call sleep as a background process
sleep_pid=${!} # get pid of the just-created background process
wait ${sleep_pid} # wait for sleep to exit
```
Which essentially creates a fork (which has its own performance downsides) and slightly higher complexity. I guess this is still faster than discovering child PIDs.

We will also need some way to find out which "primary" PID is currently running, and under some circumstances (such as when we are making a version upgrade or a new install), this information is not available externally, so we will still need to resort to some of the techniques employed by `kill_abl_pids()` which basically calls `pgrep` to get a match for certain regex, while excluding the current PID from the results. Some additional changes will likely be needed. Maybe it's all worth doing.

lynxthecat:
> > Could you be convinced?
> 
> I could, actually was considering something similar. My issue though is that the current method is basically bullet-proof. It will kill the process and any child processes, no matter what. While sending a signal to a process doesn't guarantee that the process actually exits. It will exit under normal circumstances but not, for example, while it's sleeping (it does actually receive the signals immediately when it's executing the `wait` command, from my testing). So then we will need implement all sufficiently long sleep commands like so:
> 
> ```
> sleep [X] & # call sleep as a background process
> sleep_pid=${!} # get pid of the just-created background process
> wait ${sleep_pid} # wait for sleep to exit
> ```
> 
> Which essentially creates a fork (which has its own performance downsides) and slightly higher complexity. I guess this is still faster than discovering child PIDs.

I'm not completely following this. So where we have parent trap any exit signal, then kill its children, are you concerned that the parent and/or children processes may be sleeping and thus the tear-down will be delayed? Do we have processes sleeping for a potentially longer time than we'd desire to wait before it's killed? Is that the issue?

> We will also need some way to find out which "primary" PID is currently running, and under some circumstances (such as when we are making a version upgrade or a new install), this information is not available externally, so we will still need to resort to some of the techniques employed by `kill_abl_pids()` which basically calls `pgrep` to get a match for certain regex, while excluding the current PID from the results. Some additional changes will likely be needed. Maybe it's all worth doing.

Also possible to keep a file that contains the parent PID in /var/run/adblock-lean. Aren't you doing something similar already?

adblock-lean
lynxthecat: None at all, but I think `kill -9` is POSIX compliant: https://serverfault.com/a/366485. If that matters, which it may not.

friendly-bits:
> are you concerned that the parent and/or children processes may be sleeping and thus the tear-down will be delayed?

Yes, exactly

> Do we have processes sleeping for a potentially longer time than we'd desire to wait before it's killed? Is that the issue?
Not sure - will need to check on this. But potentially even 1 sec delay may lead to errors printed in the console, for example if we send USR1 to the scheduler, the scheduler sends Kill to job, the job is sleeping for 1 sec, and in the meantime we deleted or tried to delete the temp folder it was using.

adblock-lean
friendly-bits: /proc/${pid} universally works on Linux, which I think is all we care about.

lynxthecat:
But if the scheduler as parent is handling the kills then those will happen inside the trap in the scheduler, right? So we can ensure the deletion of the folder occurs at the right moment?

friendly-bits:
> Also possible to keep a file that contains the parent PID in /var/run/adblock-lean. Aren't you doing something similar already?

This will work under most normal operating conditions, but may be an issue under some fault conditions. PID file write may fail because of no free memory, for example. Again, maybe I'm overthinking this and we can safely decide that if there is no PID file then no adblock-lean processes are running.

lynxthecat:
> > But if the scheduler as parent is handling the kills then those will happen inside the trap in the scheduler, right? So we can ensure the deletion of the folder occurs at the right moment?
> 
> The scheduler is also running asynchronously, in a background process. So if we want to ensure that everything happens at the right moment, we will need to issue kill commands and then start polling and wait till the kill is successful. So I'm not sure that we are saving any time, compared to current implementation, which normally succeeds after 1 kill attempt. The only difference is that we are not also issuing kill commands to the child processes. Under normal operating conditions, this will probably be somewhat faster. Under some fault conditions, this may be an issue. For example, if the system ran out of memory and things because very slow. In such a case I think we will want to terminate as many processes as possible and as quickly as possible, especially the processes which actually consume memory, which happen to be the grandchildren processes of the scheduler, such as gzip or sort (whose PID's the scheduler doesn't know). This may or may not be overthinking. I'm honestly not sure.

Still though, don’t you think responsible parents should look after their own children rather than expect the state to take care of them? OK the analogy doesn’t work so well given that we are talking about termination. But the point remains. I still think parents should do their own garbage collection.

friendly-bits:
> I still think parents should do their own garbage collection.

I do agree with that, and this is how it's implemented already, except when stop() is called. Following your analogy, calling stop() is like calling the police to have absolute certainty that the garbage (which in this case may be clogging the main highway in the state) is removed ASAP. The only flaw I'm seeing with this is that stop() is also called in non-emergency situations, in some cases. So maybe there should be some conditioning of when we should and shouldn't call the police inside stop(). I'll give it some thinking.

lynxthecat:
I am still trying to wrap my head around why it would not be a good idea to have parent processes keep track of any children and be responsible for killing those children before exiting themselves. So to kill the tree we kill parent, and parent kills children, before exiting. I think that's good practice in general, no?

friendly-bits:
Well, this is how it is implemented already, at least at exit. We check if the scheduler PID is still alive, if it is then we send USR1 which tells it to kill workers. My concern is that in some cases, something in the worker thread may get stuck and then killing the worker won't actually get rid of its own child processes. I'm also not sure how child processes are handled when the subshell receives a kill signal. Let's say that the job is in the middle of processing , running something like `sed|gzip|awk`, `wc` (which runs in yet another background process launched by `tee`), `sed` (which runs in yet yet another background process launched by `tee`). Will a kill signal to the job PID actually kill all that? I'm not sure.

I'm also concerned that under low memory conditions when the system becomes extremely slow, even if the kill signal propagates, this may take a very long time. So the recursive children (and grand-children and grand-grand children etc) discovery is supposed to decisively kill absolutely all the processing tree.

That said, I'm currently working on a more nuanced approach where a 'normal' kill will be tried first, then if it doesn't succeed within 3 seconds then the more radical approach is taken.

lynxthecat:
I like idea of comprehensive fallback - unleashing the terminator - if graceful fails.

friendly-bits:
So I've made some progress here, reading [this](https://stackoverflow.com/questions/8363519/how-do-i-terminate-all-the-subshell-processes) stackoverflow answer.

- Turns out there is a notion of process groups (PGID's), which (to my understanding) means "certain process and all processes it spawned".
- Signal `SIG1` can be sent to the **current process group** this way: `kill [-s SIG1] 0`
- Signal `SIG2` can be sent to **another process group** this way: `kill [-s SIG2] -[PGID]` (note the `-` before `[PGID]`
- So supposedly we could use something like `trap 'kill -s KILL 0' USR1 INT KILL TERM` in the scheduler, and this **should** effectively provide a killswitch for all processes spawned by the scheduler. 
- However this only works when job management is enabled (`set -m`), which we are currently disabling:

https://github.com/lynxthecat/adblock-lean/blob/4faec5235c5b8f6570593ca7b0d897e4322efc4e/usr/lib/adblock-lean/abl-process.sh#L532

We are disabling it in order to prevent lines like `[1]+  Done                       sleep 1` from being printed to the console when background processes we create exit.

I tested all of the above statements and found them to be true in Busybox ash. So there is a proper solution which doesn't require discovering child processes, however this will output job completion notification to the console, and what I hate about this is that they are printed not even in real time but rather when the console user enters the **next** command. This I don't know how to solve.
-

friendly-bits:
So I tested the ideas mentioned in the previous comment (spending **a lot** of time and effort), and unfortunately, it is not as easy as that. The main issue is that when starting a background process from non-interactive shell, it does not get its own process group id (`PGID`). Rather, the process is created under the parent's `PGID`. So we can not send a signal to that group without sending that signal to the main script, and because the signal needs to be `KILL` or `TERM`, this immediately kills the script, not allowing it to do cleanup. The only workaround (at least in shell code) is to use the `setsid` utility to start the process in its own session, which (presumably) will also give it its own `PGID`. However one can not call a function from `setsid`, so for that approach to work, we would need to essentially make abl-process.sh its own executable script (rather than library) and call it this way: `setsid busybox sh -c <path_to_abl-process.sh>`. There is another downside: when calling setsid, the calling process doesn't get to know the PID nor the PGID of the process it called. So we would need to implement extra methods to get that information (plus error handling when those methods fail).

While none of the above is a show-stopper, this adds complexity and is more error-prone. I'm not loving this but my **main** problem is not with that but rather with the fact that all we can do is send 1 kill signal to all processes in current process group and hope they actually receive and obey that signal. We can not verify that all the grandchild processes were actually killed because we don't know their PID's, unless we implement a similar mechanism to what is currently implemented in `kill_pids_recursive()`.

So, in conclusion, while it would be nice to have parents clean up the garbage of their children, we don't have the tools to do this right in shell code, so trying to implement this both increases complexity and reduces reliability. At this point, I think it's smarter to stick with the current code.

What do you think @lynxthecat ?

lynxthecat:
Hmm yes this is complicated. To what extent can we rely on SIGPIPE here?

friendly-bits:
I don't think there is any magical pill which would solve this problem, including SIGPIPE. I spent a lot of time on stackoverflow reading how people solve this (nobody mentioned SIGPIPE). The only solutions anyone has come up with are either child discovery (similar to what is implemented in current code) or using setsid and sending a signal to the process group. Combined with the results of my own experiments, this makes me fairly confident that there simply is no other way for a shell code implementation. Languages which have access to syscalls have some other ways to handle this but not shell.

friendly-bits:
Perhaps the current mechanism can be somewhat improved to avoid child discovery when it is not really needed (i.e. when the `stop` call was not issued by the user and no error was encountered) - I'll look into this some more.

lynxthecat:
Are you sure that terminating a process that runs a bunch of piped processes will not result in those piped processes from terminating?

friendly-bits:
> Are you sure that terminating a process that runs a bunch of piped processes will not result in those piped processes from terminating?

So here is test.sh:
```
#!/bin/sh
ttt() {
  uclient-fetch -O- https://cdimage.ubuntu.com/ubuntu-mate/releases/22.04/release/ubuntu-mate-22.04.5-desktop-amd64.iso > /dev/null
}

ttt &
sleep 1 # give the job enought time to call uclient-fetch
kill -KILL ${!}
```

call the script:
```
sh test2.sh
```

Check with top/htop: uclient-fetch is running.

lynxthecat:
Ah, but I suppose regular kill will work? 

What happens with a line involving a few piped commands? Like say the download piped through grep and cat or something?

friendly-bits:
The specific signal is irrelevant, as well as the specific commands or whether they output into a pipe or not. You are welcome to experiment with any variations.

lynxthecat:
That seems surprising to me in the sense that we are concerned with piped processes and potentially the role of SIGPIPE.

friendly-bits:
I added a few more commits addressing various issues. In particular, `kill_pids_recursive()` is now called when scheduler errors are encountered. Also fixed the issues mentioned [above](https://github.com/lynxthecat/adblock-lean/pull/133#issuecomment-2768486807).

I think we can merge now.

PR #145, opened 2025-04-08
abl-process: fix part size limiting

friendly-bits:
[This commit](418849d0ae690bd412e96f9b69a7aaac62e39bc3) broke part size limiting.

This PR implements a different method to detect hitting this limit (using return code of `grep .` to see if there are leftover bytes after limiting with `head -c`. Should not affect performance in any meaningful way.

### DIFF
--- a/usr/lib/adblock-lean/abl-process.sh
+++ b/usr/lib/adblock-lean/abl-process.sh
@@ -338,8 +338,9 @@ process_list_part()
 		ucl_err_file="${ABL_DIR}/ucl_err_${job_id}" \
 		rogue_el_file="${ABL_DIR}/rogue_el_${job_id}" \
 		list_stats_file="${ABL_DIR}/stats_${job_id}" \
+		size_exceeded_file="${ABL_DIR}/size_exceeded_${job_id}" \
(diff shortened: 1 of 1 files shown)


lynxthecat:
Is the thinking here that generating/checking existence of file would be quicker than the calculation and evaluation?

friendly-bits:
> Is the thinking here that generating/checking existence of file would be quicker than the calculation and evaluation?

No, we are calculating it anyway because we need to know the sizes later. The issue is that the calculated size is smaller than the actual size, so when if we are using the calculated size to detect whether we have hit the limit, we are not detecting that the limit was hit. Not exactly sure what's causing this, probably the fact that we are processing the data down the line and that limits the number of bytes which goes through the pipe? Anyway, without this patch, part size limit is not detected - with this patch, it is.

friendly-bits:
Ah, I know what's causing this: the `wc` command runs after we do some processing, while the `head -c` command runs before the processing. Which makes sense because we don't want to download infinite amount of data. But the shortcoming is that `wc` sees registers less bytes than max limit.

PR #146, opened 2025-04-09
Some improvements

friendly-bits:
- In current code, final size limiting is implemented in a non-optimal way:
```
{ head -c "${max_blocklist_file_size_B}"; head -c 1 > "${ABL_DIR}/abl-too-big.tmp"; cat 1>/dev/null; }
```

As I discovered recently, when 3 commands are subjected to input pipe (in this case `head`, `head` and `cat`), the 3rd command may not see the input stream. Also there is no need in 3 commands here. This PR implements a better approach to size limiting with `grep . >/dev/null && touch "${ABL_DIR}/abl-too-big.tmp"`

- With current code, in some cases `stop` is called twice under fault conditions. This PR fixes that.

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -1231,7 +1231,7 @@ stop()
 	done
 	msg="${msg% }"
 
-	init_command stop || exit 1
+	init_command stop || { FAIL_STOP_REQ=''; exit 1; }
 
 	# get DNSMASQ_CONF_D without calling load_config()
 	[ -f "${ABL_CONFIG_FILE}" ] &&
@@ -1247,6 +1247,7 @@ stop()
 	restart_dnsmasq -nostop || stop_rc=1
 	log_msg -purple "" "Stopped adblock-lean."
 	[ -n "$noexit" ] && return "${stop_rc}"
+	FAIL_STOP_REQ=
 	exit "${stop_rc}"
 }
 
--- a/usr/lib/adblock-lean/abl-process.sh
+++ b/usr/lib/adblock-lean/abl-process.sh
@@ -794,7 +794,7 @@ gen_and_process_blocklist()
 	tee >(wc -c > "${ABL_DIR}/final_list_bytes") |
 
 	# limit size
-	{ head -c "${max_blocklist_file_size_B}"; head -c 1 > "${ABL_DIR}/abl-too-big.tmp"; cat 1>/dev/null; } |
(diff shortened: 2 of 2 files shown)


lynxthecat:
Seems great. I suppose head doesn’t give a useful exit status?

friendly-bits:
The idea here is to sink the remaining data to /dev/null (of we don't sink it and there is a lot of remaining data, this extra data causes a huge slowdown in execution), while registering its existence via a temp file. I'm not sure where the data goes if we use timed `read` command (as you did not specify a variable to read into - is this a valid syntax?) but even if the data is discarded for the first 0 seconds, there will be still more data after that, and I'm not sure what will happen to that data. So there are a lot of unknowns with `read -t 0` while the only benefit is avoiding a fairly cheap binary call to grep. I think that `grep . >/dev/null` is better suited for this task.

friendly-bits:
I tested this on my Linux Mint installation:
```
cat abl-install.sh | { head -c500000; read -t 0 && touch bruh; cat > /dev/null; }
```

Unfortunately, the file 'bruh' is created regardless of whether there is extra data in the pipe or not.

lynxthecat:
Argh. Why? What's doing on here?

friendly-bits:
Apparently, `read` always returns 0?

lynxthecat:
> Apparently, `read` always returns 0?

I'm trying to reconcile what's happening here with:

```
		/* "If timeout is 0, read returns immediately, without trying
		 * to read any data. The exit status is 0 if input is available
		 * on the specified file descriptor, non-zero otherwise."
```

It seems maybe EOF from before the PIPE results in file descriptor still being considered to have input available.

lynxthecat:
I was chatting with @jaggzh on IRC and it seems that our discussion inspired him to write:

https://github.com/jaggzh/iny

albeit not sure that we can leverage that in our project.

friendly-bits:
> https://github.com/jaggzh/iny

This seems like a much needed thingy but unless it's included in the OpenWrt repo, we can not benefit from it.

friendly-bits:
> I'm trying to reconcile with:

Well, apparently this doesn't work in practice this way.

lynxthecat:
How about: 

`read -rn1 -d ‘’`

friendly-bits:
> read -rn1 -d ‘’

This actually does work, and is slightly faster than the grep method over 1000 iterations (up to 10% faster).

friendly-bits:
(PR updated to incorporate your idea)

PR #164, opened 2025-05-15
Improve update

lynxthecat:


friendly-bits:
> How did you manage to integrate these into #165?

The cause of the issue was that I wanted to be able to test upgrading v0.7.1 to v0.7.2 in my fork. However, the code in v0.7.1 doesn't have a provision for this. So I added the commit 0fe527fc46340d50a218dec4a7db10d3d3391ddf to the `master` branch in my fork. However, as this was an afterthought, the `improve_update` branch was already many commits ahead, and this means that the histories of the two branches in my fork diverged. While testing, I needed to be able to merge the `improve_update` branch into the `master` branch in my fork, which was now impossible without re-basing, and re-basing was too cumbersome at that point. So I simply made copies of all changed files, recreated the `improve_update` branch in my fork, and pasted the copied versions and made the huge commit 3307d4d11e9375c9580df0fbd6bf9c751f37096a which incorporated all the changes.

PR #165, opened 2025-05-17
Improve update and other changes

friendly-bits:
- Rather than writing the version string into each of the installed files, only write it into the main file
- Rather than checking libraries by comparing version strings, create a file with md5sums of all installed files and check against that. Together with the above bullet point, this is both more robust and allows to minimize the number of flash writes when updating adblock-lean scripts. Currently all scripts are re-written upon update because the version string changes every time in all of them. This way only the main script will be re-written unconditionally and all others will be only re-written when they actually change.
- Because we discovered that GitHub limits the number of API requests, implement local API query caching which works around that
- Support a new update channel: `branch=<gh_branch>`
- Rather than registering the version string like so:
   ```
   # ABL_VERSION=<upd_channel>_<version>
   ```
   use 2 variables:
   ```
   ABL_VERSION="<version>"
   ABL_UPD_CHANNEL="<upd_channel>"
   ```
- Remove some unused functions

This also makes it possible to override the GH repo author by setting the variable `ABL_REPO_AUTHOR`. I added this in order to be able to test upgrading more easily.

This PR solves some bugs present in #164. Unfortunately, at some point I had to reset the commit history in my branch, so I can not easily add the extra bugfixes to #164. Hence this separate PR.

Extra changes:

- Implement support for additional compression utils: GNU gzip, pigz, zstd
- Support the `-y` option to auto-approve all suggested changes during adblock-lean version updates

### DIFF
--- a/README.md
+++ b/README.md
@@ -298,7 +298,7 @@ blocklist_urls="https://raw.githubusercontent.com/hagezi/dns-blocklists/main/wil
 ### Blocklist compression
 - By default, adblock-lean uses compression while processing downloaded blocklist/allowlist parts, and compresses the final blocklist before loading it into dnsmasq. This helps to reduce memory use.
 - Supported compression utilities: Busybox gzip (every OpenWrt system has this built-in), GNU gzip, pigz and zstd. The latter two utilities support multithreaded compression.
-- adblock-lean automatically sets parameters for any of the supported utilities for reasonable balance between speed and memory usage. You can specify your preferred compression utility in the `compression_util` option (default is `gzip`). You can also specify parameters to pass to the compression utility, separately for intermediate compression (`intermediate_compression_options`) and for the final blocklist compression (`final_compression_options`).
+- adblock-lean automatically sets parameters for any of the supported utilities to provide a reasonable balance between speed and memory usage. You can specify your preferred compression utility in the `compression_util` option (default is `gzip`). You can also specify parameters to pass to the compression utility, separately for intermediate compression (`intermediate_compression_options`) and for the final blocklist compression (`final_compression_options`).
 - Final blocklist compression depends on appropriate addnmount entries existing in `/etc/config/dhcp`. After changing the compression utility in the config file, make sure to run `service adblock-lean setup` in order to update the addnmount entries (answer `e` when asked whether to create new config or use existing config).
 
 ## Whitelist mode
--- a/README.md
+++ b/README.md
@@ -297,7 +297,7 @@ blocklist_urls="https://raw.githubusercontent.com/hagezi/dns-blocklists/main/wil
 
 ### Blocklist compression
 - By default, adblock-lean uses compression while processing downloaded blocklist/allowlist parts, and compresses the final blocklist before loading it into dnsmasq. This helps to reduce memory use.
-- Supported compression utilities: Busybox gzip (every OpenWrt system has this built-in), GNU gzip, pigz and zstd. The latter 2 utilities support multithreaded compression.
+- Supported compression utilities: Busybox gzip (every OpenWrt system has this built-in), GNU gzip, pigz and zstd. The latter two utilities support multithreaded compression.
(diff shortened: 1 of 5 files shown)


friendly-bits:
I added a few more commits implementing some feature requests and fixing a bug in the update mechanism. I think the code is ready to be merged. Note that because of some workarounds employed by this version upgrade, we need to make a release (I suggest v0.7.2), otherwise it'll be impossible to update to this version (it will automatically pull v0.7.1 and install it instead).

README.md
```diff
@@ -291,6 +295,12 @@ blocklist_urls="https://raw.githubusercontent.com/hagezi/dns-blocklists/main/wil
 blocklist_urls="https://raw.githubusercontent.com/hagezi/dns-blocklists/main/wildcard/pro-onlydomains.txt https://raw.githubusercontent.com/hagezi/dns-blocklists/main/wildcard/tif-onlydomains.txt"
 ```
 
+### Blocklist compression
+- By default, adblock-lean uses compression while processing downloaded blocklist/allowlist parts, and compresses the final blocklist before loading it into dnsmasq. This helps to reduce memory use.
+- Supported compression utilities: Busybox gzip (every OpenWrt system has this built-in), GNU gzip, pigz and zstd. The latter 2 utilities support multithreaded compression.
```
lynxthecat, line 300:
```suggestion
- Supported compression utilities: Busybox gzip (every OpenWrt system has this built-in), GNU gzip, pigz and zstd. The latter two utilities support multithreaded compression.
```

README.md
```diff
@@ -291,6 +295,12 @@ blocklist_urls="https://raw.githubusercontent.com/hagezi/dns-blocklists/main/wil
...
+- adblock-lean automatically sets parameters for any of the supported utilities for reasonable balance between speed and memory usage. You can specify your preferred compression utility in the `compression_util` option (default is `gzip`). You can also specify parameters to pass to the compression utility, separately for intermediate compression (`intermediate_compression_options`) and for the final blocklist compression (`final_compression_options`).
```
lynxthecat, line 301:
```suggestion
- adblock-lean automatically sets parameters for any of the supported utilities to provide a reasonable balance between speed and memory usage. You can specify your preferred compression utility in the `compression_util` option (default is `gzip`). You can also specify parameters to pass to the compression utility, separately for intermediate compression (`intermediate_compression_options`) and for the final blocklist compression (`final_compression_options`).
```

README.md
```diff
@@ -406,6 +416,13 @@ After updating adblock-lean, run the command:
 service adblock-lean start
 ```
 
+## Advanced version update options
+
+adblock-lean implements a flexible update system which supports following options (use with `service adblock-lean update`):
+- `-y` : pre-approve any configuration changes suggested by the version update mechanism and automatically start the updated adblock-lean version
+- `-f` : update without calling the `stop` command first and do not load adblock-lean library scripts - this is mainly useful to fix a broken installation
+- `-v < [version]|[update_channel]|commit=[commit_hash] >` : either install a specific adblock-lean version (for example `-v 0.7.2`); or specify an update channel (either `-v release` or `-v snapshot` or `-v branch=<branch_name>`) - this will install the latest version from the corresponding update channel and change the version update behavior so it checks in that update channel for future updates; or install a version corresponding to a specific commit to the `master` branch. These options are mainly helpful for testing. Most users should be fine with the default update behavior which is following the `release` update channel.
```
lynxthecat, line 424:
```suggestion
- `-v < [version]|[update_channel]|commit=[commit_hash] >` : either install a specific adblock-lean version (for example `-v 0.7.2`); or specify an update channel (either `-v release` or `-v snapshot` or `-v branch=<branch_name>`) - this will install the latest version from the corresponding update channel and change the version update behavior so it checks in that update channel for future updates; or install a version corresponding to a specific commit to the `master` branch. 

These options are mainly helpful for testing. Most users should be fine with the default update behaviour which follows the `release` update channel.
```

lynxthecat:
Is this otherwise ready to merge do you think?

friendly-bits:
Above commit fixes some logic issues. Now it's ready to merge.

PR #181, opened 2025-06-03
Various improvements

friendly-bits:
This PR is mainly about code maintenance and bugfixes.

In particular, this fixes a bug with files `.abl-extract_blocklist`, `abl-conf-script` not removed when adblock-lean is paused.
Also this adds the `.gitattributes` file as discussed in #93.

### DIFF
--- a/adblock-lean
+++ b/adblock-lean
@@ -431,7 +431,7 @@ detect_util()
 # 1 (optional) : '-f' to force re-detection
 # sets $AWK_CMD, $SED_CMD, $SORT_CMD
 # shellcheck disable=SC2120
-detect_utils()
+detect_main_utils()
 {
 	[ -n "${MAIN_UTILS_DETECTED}" ] && [ "${1}" != '-f' ] && return 0
 	unset SED_CMD AWK_CMD SORT_CMD
@@ -1982,7 +1982,7 @@ then
 fi
 rm -f /tmp/abl-test
 
-detect_utils || graceful_exit 1
+detect_main_utils || graceful_exit 1
 
 # register 1st arg as $ABL_CMD when called by 'sh /etc/init.d/adblock-lean'
 if [ "${0}" = "${ABL_SERVICE_PATH}" ] && [ -n "${1}" ]
--- a/adblock-lean
+++ b/adblock-lean
@@ -1565,7 +1565,7 @@ start()
 		sleep "${random_delay_mins}m"
 	fi
 
-	try_export_existing_blocklist
+	try_export_blocklist
 	[ ${?} = 1 ] && exit 1
 
 	FAIL_STOP_REQ=1
@@ -1702,7 +1702,7 @@ pause()
 	esac
 	FAIL_STOP_REQ=1
 	reg_action -purple "Pausing adblock-lean." || exit 1
-	try_export_existing_blocklist || exit 1
+	try_export_blocklist || exit 1
 	restart_dnsmasq || exit 1
 	FAIL_STOP_REQ=
 	log_msg -purple "adblock-lean is now paused."
--- a/adblock-lean
+++ b/adblock-lean
@@ -827,12 +827,16 @@ get_abl_run_state()
 {
 	[ -n "${DNSMASQ_CONF_D}" ] || { reg_failure "\$DNSMASQ_CONF_D is not set"; return 1; }
 	local f
-	for f in "${DNSMASQ_CONF_D}/.abl-blocklist${COMPR_EXT}" "${DNSMASQ_CONF_D}/abl-blocklist"
+	for f in "${DNSMASQ_CONF_D}/.abl-blocklist"* "${DNSMASQ_CONF_D}/abl-blocklist"
 	do
-		[ -f "${f}" ] && return 0
+		[ -n "${f}" ] && [ -f "${f}" ] && return 0
 	done
 
-	[ -f "${ABL_DIR}/prev_blocklist${COMPR_EXT}" ] && return 3
+	for f in "${ABL_DIR}/prev_blocklist"*
+	do
+		[ -n "${f}" ] && [ -f "${f}" ] && return 3
+	done
+
 	return 4
 }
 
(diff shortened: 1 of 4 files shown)


adblock-lean
```diff
@@ -357,50 +357,66 @@ rc_enabled()
...
-		res_cmd='' res_name='' _util _util_path='' busybox_variant_avail=''
+		res_cmd='' res_name='' _util _util_path='' busybox_variant_avail='' skip_links='' path_exists='' path_is_symlink='' \
+		opt me=detect_util
 
-	are_var_names_safe "${du_out_var}" || return 1
+	[ ${#} -ge 4 ] || { reg_failure "${me}: not enough arguments"; return 1; }
+	shift 4
 
-	[ "${5}" = '-b' ] && busybox_variant_avail=1
+	while getopts :bs opt
```
lynxthecat, line 369: What does getopts do?

friendly-bits: Basically parses arguments.

See [here](https://forum.openwrt.org/t/optimization-of-adblock-lean/203108/1749).

lynxthecat: Ah, nice. Clever usage (as always).

friendly-bits: I've become accustomed to using getopts while working on geoio-shell as geoio-shell expects to be configured via command-line arguments, rather than via the config file. So tons of arguments parsing code. For an especially egregious example:
https://github.com/friendly-bits/geoip-shell/blob/d59d4f07d08b0df66cd2fb92059d05f523d85949/geoip-shell-manage.sh#L163

lynxthecat: Wow - I see what you mean.

adblock-lean
```diff
@@ -812,12 +827,16 @@ get_abl_run_state()
...
-	for f in "${DNSMASQ_CONF_D}/.abl-blocklist${COMPR_EXT}" "${DNSMASQ_CONF_D}/abl-blocklist"
+	for f in "${DNSMASQ_CONF_D}/.abl-blocklist"* "${DNSMASQ_CONF_D}/abl-blocklist"
+	do
+		[ -n "${f}" ] && [ -f "${f}" ] && return 0
+	done
+
+	for f in "${ABL_DIR}/prev_blocklist"*
 	do
-		[ -f "${f}" ] && return 0
+		[ -n "${f}" ] && [ -f "${f}" ] && return 3
```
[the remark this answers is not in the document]
friendly-bits: When using wildcards in a `for` loop to check for the existence of files, when no matching files are found, something unexpected happens: the variable is assigned the literal string including the wildcard. So for example:
```
for file in /bogus_dir/*
do
  echo "file is '$file'"
done
```

Output:
```
file is '/bogus_dir/*'
```

To contract that, you have to check that the file actually exists, otherwise skip.

There is another issue: when setting IFS to newline and looping over newline-separated list, if empty lines are encountered, the variable will be set to empty string. So ideally we also want to check for that condition and skip if it's true. In this particular part of code, this should never happen, but it doesn't hurt to have the check in place and this may prevent future bugs when the code is modified.

usr/lib/adblock-lean/abl-lib.sh
```diff
@@ -324,27 +324,31 @@ mk_preset_arrays()
...
 # 1 - mini|small|medium|large|large_relaxed
 # 2 - (optional) '-d' to print the description
 # 2 - (optional) '-n' to print nothing (only assign values to vars)
-gen_preset()
+set_preset_vars()
 {
-	# rounds down to reasonable number of digits, based on the input number of digits
+	# keeps first two digits, replaces others with 0's
 	# 1 - var for I/O
 	reasonable_round()
```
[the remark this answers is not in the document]
friendly-bits: That's a semantics question. Technically, the modified code here does the exact same thing as the code before modification, except in a more straightforward, universal and reliable way and it doesn't have the bug which the pro's revision had, which is when the input had 2 digits, the output was empty string. I updated the description to say exactly what's being done. I think this does qualify to rounding, as we are keeping the useful n most significant digits and discard m least significant digits. After some experimentation, I found the value of n fitting this specific application to be 2.

friendly-bits:
Hi @lynxthecat, reminding about this PR.

lynxthecat:
At this line here:

https://github.com/lynxthecat/adblock-lean/blob/1159a3b3bc15cdb4d9a0659b467a2f0f00988c61/usr/lib/adblock-lean/abl-lib.sh#L560

Should we add a comment along the lines: "Unable to suggest a suitable preset. Please pick your own."?

friendly-bits:
> At this line here:
> 
> https://github.com/lynxthecat/adblock-lean/blob/1159a3b3bc15cdb4d9a0659b467a2f0f00988c61/usr/lib/adblock-lean/abl-lib.sh#L560
> 
> Should we add a comment along the lines: "Unable to suggest a suitable preset. Please pick your own."?

`${preset}` will be empty string if `get_def_preset()` failed, most likely because we failed to detect total memory capacity of the system. In this case, `get_def_preset()` prints an error message:

https://github.com/lynxthecat/adblock-lean/blob/1159a3b3bc15cdb4d9a0659b467a2f0f00988c61/usr/lib/adblock-lean/abl-lib.sh#L634

I thought this was sufficient and the user will understand why they are asked to pick a preset. If you think we should add an extra message, no problem.

lynxthecat:
Yeah I understood that this would occur at least on the failure to detect memory. I mean perhaps it's obvious enough, but I thought perhaps it might be reassuring that the consequence is not dire and only amounts to inability to suggest suitable preset. Just a suggestion from me - I'll leave it with you whether to add or leave as is.

I didn't have any other thoughts or suggestions.

friendly-bits:
All right, I added a log message saying approximately what you suggested (the phrasing is a bit different because this may also occur when `setup()` is called by the luci app and we need to accommodate that case).



> I don't have any other thoughts or suggestions.

Do you mean the whole PR? If so, should we merge?

lynxthecat:
> All right, I added a log message saying approximately what you suggested (the phrasing is a bit different because this may also occur when setup() is called by the luci app and we need to accommodate that case).

Excellent - I like that better than my own suggestion.

> Do you mean the whole PR? If so, should we merge?

Yes I think so!

Issue #151, opened 2025-05-02
Configuration Ideas using the Openwrt UCI System


languagegame:
Great program.

Here are some ideas to make the program more integrated/dynamic with the rest of Openwrt.  Some of these ideas may be easier to implement than others, so this is more a wishlist

**1) Append `option confdir` to the selected dnsmasq instance for ad blocking (if such option does not already exist)**
During the current setup process, the code pulls dnsmasq configuration using `/etc/init.d/dnsmasq info` and then parses it for things like the list of dnsmasq instances, confdir folders and names. 
Here is an alternate way that likely requires much less code and would have some other benefits.  Use /etc/config/dhcp itself to parse the configuration of dnsmasq.  This would require that the confdir option be set explicitly for the selected dnsmasq instance for adblocking, as I don't believe Openwrt sets this option by default on the initial installation.  So the setup code would parse /etc/config/dhcp for the list of dndmsasq instances, and once the user selects the instance(s) for ad blocking, the setup program would append an option with the confdir like:
`uci set dhcp.@dnsmasq[0].confdir='/tmp/dnsmasq.adblock-lean.d'`
`uci commit`
This has many possible advantages:
- The user doesn't have to run adblock-lean setup again if they change the confdir later in /etc/config/dhcp
- Fewer configuration values for adblock-lean are needed.  The program now can just use DNSMASQ_INDEX to get the confdir later.  It does not need to store DNSMASQ_INSTANCE or DNSMASQ_CONF_D, so two less configuration values needed.
- Eliminates a bunch of JSON parsing code from adblock-lean, as the current code has to parse through the JSON output from /etc/init.d/dnsmasq info` whereas now the dnsmasq configuration can all be simply read from the UCI configuration
- Avoids possible race condition.  By having adblock-lean read the dnsmasq configuration from the Openwrt UCI system, adblock-lean could potentially download and store the blocklists in the right conf-dir even before dnsmasq is initialized

If there is a concern about appending another option to the dhcp configuration file in the Openwrt UCI Configuration System, adblock-lean is already appending an option for addnmount to enable blocklist decompression.  So it would be adding 2 options to dhcp instead of just addnmount.
If there is a concern about changing the default confdir folder for the selected instance of dnsmasq, the default folder is already rather arbitrary so might as well be declared explicitly

**2) Allow user to select multiple dnsmasq instances for adblocking.**
[...]

PR #183, opened 2025-06-24
Support adblocking on multiple instances

friendly-bits:
This PR implements support for adblocking on multiple dnsmasq instances (#151), and in addition:
- select_dnsmasq_instance(): automate creating addnmounts
- improve the config parser
- implement automatic migration of config options
- change some config options
- use a separate directory for temp files created in processing
- fix error message for missing compression utilities
- code cleanup


lynxthecat:
@friendly-bits is it anticipated that the blocklist files will be the same or different for the multiple instances? It seems to me that if the same I wonder if some memory could be saved by saving just the one common copy then loading in that copy for the multiple instances rather than saving multiple copies. However this would only work where the blocklist files were indeed the same. So it might not be a great idea. Just thinking out loud.

friendly-bits:
> is it anticipated that the blocklist files will be the same or different for the multiple instances?

Currently the same. We could implement support for different blocklists but that would require very substantial code changes and re-thinking of the config structure (either switching to UCI config or perhaps splitting the config into general config file and then a separate config file for each of the different blocklists).



> some memory could be saved by saving just the one common copy then loading in that copy for the multiple instances rather than saving multiple copies.

This is already implemented. Only one copy is generated during processing. When multiple instances are configured, current implementation moves the blocklist file to `${SHARED_BLOCKLIST_PATH}` (/var/run/adblock-lean/abl-blocklist[.gz]) and points to this path in each of the `.abl-extract_blocklist` files (which are created in the relevant dnsmasq config directories for each of the configured instances). addnmount entries creation and validation logic handles all this as well.

The only memory waste is caused by the fact that each dnsmasq instance ultimately loads same blocklist into its own memory. I think this could be mitigated by having a more elaborate dns resolution setup where you have 1 instance serving dns resolution (which you load the blocklist into), and then any number of extra instances serving dhcp or whatever and forwarding dns requests to the resolving instance. I don't know enough about dns resolution to implement something like that programmatically (and reliably handle all possible scenarios) but I believe the users should be able to create such setups manually, and current adblock-lean implementation should work with them just fine as-is.

lynxthecat:
Ah terrific @friendly-bits - from my perspective this means that you were already one step ahead of me and this optimization has already been implemented.

PR #191, opened 2025-07-08
Implement calculate limits and other improvements

friendly-bits:
This PR implements support for command `calculate_values`, as discussed in the forum.

Edit: command name changed to `calculate_limits`.

Edit 2:

added a few more commits:
- Fix converting short ID to URL for dnsmasq format
- Fix packing entries with awk (turns out we had a regression which caused only sed to be used for packing)
- process_list_part(): refactor for better readability and to avoid unnecessary logic inside the processing pipeline
- process_list_part(): delay reading processed list stats to when they are actually needed (may avoid the need to `sleep 1` when the filesystem did not have time to prepare the file yet)
- get_gh_ref(): minor refactor to work around GH syntax highlighting bug
- gen_list_parts(): avoid unnecessary I/O
- parse_config(): handle duplicate keys

### DIFF
--- a/usr/lib/adblock-lean/abl-lib.sh
+++ b/usr/lib/adblock-lean/abl-lib.sh
@@ -381,17 +381,17 @@ do_setup()
 mk_preset_arrays()
 {
 	# quasi-arrays for presets
-	# cnt - target elements count/1000, mem - memory in MB
+	# urls_cnt - urls count, cnt - target elements count, mem - memory in MB
 	mini_urls="hagezi:pro.mini" \
-		mini_cnt=85 mini_mem=64
+		mini_urls_cnt=1 mini_cnt=85000 mini_mem=64
 	small_urls="hagezi:pro" \
-		small_cnt=250 small_mem=128
+		small_urls_cnt=1 small_cnt=250000 small_mem=128
 	medium_urls="hagezi:pro hagezi:tif.mini" \
-		medium_cnt=350 medium_mem=256
+		medium_urls_cnt=2 medium_cnt=350000 medium_mem=256
 	large_urls="hagezi:pro hagezi:tif" \
-		large_cnt=1200 large_mem=512
+		large_urls_cnt=2 large_cnt=1200000 large_mem=512
 	large_relaxed_urls="hagezi:pro hagezi:tif" \
-		large_relaxed_cnt=1200 large_relaxed_mem=1024 large_relaxed_coeff=2
+		large_relaxed_urls_cnt=2 large_relaxed_cnt=1200000 large_relaxed_mem=1024 large_relaxed_coeff=2
 }
 
 # sets $blocklist_urls, $min_good_line_count, $max_blocklist_file_size_KB, $max_file_part_size_KB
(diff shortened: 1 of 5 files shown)


lynxthecat:
Looks good to me. One suggestion is to replace ‘calculate_values’ with ‘suggest_mem_values’, as I think that’s a more informative command name. Does that seem reasonable?

friendly-bits:
I agree that calculate_values does not fully describe what the command is doing. On the other hand, neither does calculate_mem_values. And the latter sounds kinda odd. I think that probably there is no laconic combination of words which would be completely descriptive of what the command does, so the user will have to read the actual description of the command anyway.

lynxthecat:
calculate_values jars for me because I think it’s too vague. Which values? I’m sure this could be improved upon one way or another.

friendly-bits:
Maybe calculate_limits then?

friendly-bits:
While the limits are indirectly related to memory use, I wouldn't call them memory limits. Especially min_good_line_count. So I think calculate_limits it is.

lynxthecat:
And still better than calculate_safety_limits? What do you think? Certainly calculate_limits is more concise.

friendly-bits:
Safety is way too strong of a word for this, IMO. And yeah, IMO conciseness is important.

friendly-bits:
All right, I changed the command name to calculate_limits. Do you think we can merge this?

lynxthecat:
Yes absolutely. Nice addition!

friendly-bits:
I added a few more commits:
- Fix packing entries with awk (turns out we had a regression which caused only sed to be used for packing)
- process_list_part(): refactor for better readability and to avoid unnecessary logic inside the processing pipeline
- process_list_part(): delay reading processed list stats to when they are actually needed (may avoid the need to `sleep 1` when the filesystem did not have time to prepare the file yet)
- get_gh_ref(): minor refactor to work around GH syntax highlighting bug

usr/lib/adblock-lean/abl-lib.sh
```diff
@@ -421,43 +449,87 @@ set_preset_vars()
...
+			read -r urls_cnt
+			case "${urls_cnt}" in ''|*[!0-9]*)
+				print_msg "Invalid input '${urls_cnt}'. Please enter a number."
+				continue
+			esac
+			break
+		done
+	else
+		case "${tgt_entries_cnt}" in ''|*[!0-9]*)
+			reg_failure "calculate_limits: invalid entries count '${tgt_entries_cnt}'."; return 1 ;;
```
lynxthecat, line 487:
```suggestion
			reg_failure "calculate_limits: Invalid entries count '${tgt_entries_cnt}'."; return 1 ;;
```

usr/lib/adblock-lean/abl-lib.sh
```diff
@@ -421,43 +449,87 @@ set_preset_vars()
...
 	# Default values calculation:
 	# Values are rounded down to reasonable degree
 
 	final_entry_size_B=20 # assumption
 	source_entry_size_B=20 # assumption for raw domains format. dnsmasq source format not used by default
 
-	# target_lines_cnt / 3.5
-	min_good_line_count=$((tgt_lines_cnt_k*10000/35))
+	# tgt_entries_cnt / 3.5
+	min_good_line_count=$((tgt_entries_cnt*10/35))
```
lynxthecat, line 506: Where does the 10/35 come from here?

friendly-bits: Oops, please ignore the first part of my comment. I got confused looking at the pre-commit version of the code. We do have the variable `tgt_entries_cnt` now :)

lynxthecat:
Looks good to me. Great to see more case statements!

PR #202, opened 2025-10-11
Various improvements

friendly-bits:
- implement get_blocklist_timestamp()
- remove blocklist backup file on stop() - fixes stop from paused state
- improve console messages
- work around GH syntax highlighting bug
- test_url_domains: fix connectivity check when short id's are used
- add support for hosts lists
- add support for stevenblack short list id's
- change config options
- abl-process: print short id rather than url
- align data in processing messages when possible
- fix occasional invalid PID in $RUNNING_PIDS
- parse_config: fix a bug with migration variables generation
- parse_config: fix strings printed to console when config fixes are required
- reduce the number of messages sent to system log to minimum
- send all messages to syslog when $ABL_DEBUG_LOG is set
- improve color highlighting in progress messages
- add support for Hagezi social list
- implement support for download mirror selection
- add mirrors for Hagezi, oisd, stevenblack
- minor logic improvements in the config parser
- abl-install: avoid dialogs when calling stop


adblock-lean
```diff
@@ -1705,13 +1743,16 @@ status()
 				{ reg_failure "No entries found in the blocklist file."; run_state=1; }
 			luci_good_line_count=${active_entries_cnt}
 			int2human active_entries_cnt_human "${active_entries_cnt}"
-			log_msg "The dnsmasq check passed and the presently installed blocklist has entries count: ${active_entries_cnt_human}."
-			log_msg -green "" "adblock-lean is active."
+			print_msg -green "" \
+				"The dnsmasq check passed and the presently installed blocklist has entries count: ${blue}${active_entries_cnt_human}" \
+				"adblock-lean is active."
+			timestamp="$(get_blocklist_timestamp -h)" &&
+				print_msg "" "Blocklist updated on: ${blue}${timestamp}${n_c}"
```
lynxthecat:
```suggestion
				print_msg "" "Blocklist was updated on: ${blue}${timestamp}${n_c}"
```

adblock-lean
```diff
@@ -1705,13 +1743,16 @@ status()
...
 			gen_stats -noexit
 		else
 			reg_failure "The dnsmasq check failed with existing blocklist file."
 			run_state=1
-			log_msg "Consider a full reset by running: 'service adblock-lean restart'."
+			print_msg "Consider a full reset by running: 'service adblock-lean restart'."
```
lynxthecat: What is meant by here by a full reset? Why does a restart achieve that?

adblock-lean
```diff
@@ -1489,6 +1501,31 @@ get_config_format()
 	fi
 }
 
+# 1 - <-s|-h> to output as [s]econds since epoch or [h]uman-readable
+get_blocklist_timestamp()
```
lynxthecat, line 1510: Nice idea.

usr/lib/adblock-lean/abl-lib.sh
```diff
@@ -12,6 +12,19 @@ RECOMMENDED_PKGS="gawk sed coreutils-sort"
 RECOMMENDED_UTILS="awk sed sort"
 ABL_CRON_SVC_PATH=/etc/init.d/cron
 ALL_PRESETS="mini small medium large large_relaxed"
+ALL_LIST_FORMATS="raw dnsmasq hosts"
+
+OISD_DL_URL="oisd.nl"
+OISD_LISTS="big small nsfw nsfw-small"
+
+HAGEZI_DL_URL="https://raw.githubusercontent.com/hagezi/dns-blocklists/main"
+HAGEZI_LISTS="anti.piracy blocklist-referral doh doh-vpn-proxy-bypass dyndns fake gambling gambling.medium gambling.mini hoster \
```
lynxthecat: The NSFW angle is interesting to me because it goes beyond advert blocking and into the area of protection for children or adults wanting to filter out bad content. Perhaps we should advertise the fact that adblock-lean can actually be used for family or NSFW filtering.

adblock-lean
friendly-bits: I believe this message predates my contributions to adblock-lean :) Overall it makes sense to me. If adblock-lean is not stopped or paused but adblocking is not working then possibly something weird happened and adblock-lean restart might fix that. I suppose the message could be rephrased as `Consider running 'service adblock-lean restart` in order to restore adblocking.`

adblock-lean
friendly-bits: This was originally intended as a facility for luci-app-adblock-lean (which implements timestamp check but it doesn't account for adblocking on multiple dnsmasq instances). Then I thought that printing the timestamp in the status also makes sense.

usr/lib/adblock-lean/abl-lib.sh
friendly-bits: Sure, would you like to update the README?

friendly-bits:
The 3 commits above fix a couple of issues I discovered while reviewing the code again, and re-phrase the message you asked about. Do you have other concerns with the code or is this good to merge?

adblock-lean
lynxthecat: Yes I agree with the change in wording - I think the reset language is confusing.

usr/lib/adblock-lean/abl-lib.sh
lynxthecat: Yes, when I get a chance, but I can do this after merge in I think.

lynxthecat:
I think this is a fantastic update and good to merge.

PR #203, opened 2025-11-05
Enhance README with family filtering information

lynxthecat:
Added family filtering section to README.md to provide guidance on blocking adult content.


friendly-bits:
I think that some people may not like this use case [of enforcing content blocking]. Whether enforced content blocking is good or bad is a matter of opinion and I don't think we need to take a stand on this or suggest specific blocking practices. So I would rephrase the first inserted paragraph in a more neutral way:

```
Besides blocking ads, adblock-lean can also be leveraged to block unwanted content in general. For example, to block websites delivering malware or adult content.
```

And then shorten and slightly rephrase the second inserted paragraph:

```
Those interested in blocking adult content may want to consider setting the main DNS resolver to Cloudflare's [1.1.1.1 for Families](https://blog.cloudflare.com/introducing-1-1-1-1-for-families/), and then selectively blocking any further unwanted content with the help of adblock-lean.
```

PR #209, opened 2026-02-03
Next version

friendly-bits:
**Major changes**:
- Use the term `blockset` (rather than `blocklist`) when referring to the final blocklist, in order to avoid confusion between that and blockset parts of type `blocklist`
- Implement support for persistent blockset
- Implement hotplug script functionality in order to improve robustness of persistent blocksets (currently untested/unfinished, disabled in the code)
- Implement support for multiple blocksets
- Very significant changes to the config subsystem required to support multiple blocksets
- Add actions: `create_addnmounts`, `gen_global_config`, `list_blockset_configs`, `gen_blockset_config`, `rm_blockset_config`, `gen_persist_blockset`
- Add the abl-manage.sh library script
- Rework messaging/logging system
- More robust error checking, especially in the processing pipeline
- Change conf-script implementation: now only one conf-script is created (rather than two)
- Change addmounts support: potentially unsafe `/bin/busybox` addnmount is no longer required or created. Instead, addnmounts for `[path-to-cat]`, `[path-to-gzip]`, `[path-to-logger]` are created, depending on user config
- Change dnsmasq support implementation: now only required dnsmasq instances are stopped or restarted (rather than all dnsmasq instances)
- A lot of refactoring and logic changes
- Reduce the default value of boot_start_delay from 120s to 30s
- Remove support for the other compression utilities besides gzip, and remove options to call the compression utility with custom arguments (maybe reimplement this functionality later)
- Lots of changes to console messages

**Bugfixes**:
- Fix stopping another adblock-lean instance when it's sleeping
- get_uptime_ms(): fix fraction digits normalization for digits > 3
- Avoid calling exit when the init script is not called with a specific action - prevents unexpected exits when the script is sourced, e.g. by rc.common
- Fix processing addnmount entries which include whitespaces
- Other minor fixes

**Other improvements**:
- Accept any valid action when called via `sh /etc/init.d/adblock-lean`
- Move process substitution testing code into abl-install.sh (rather than executing it every time the the init script is called)
- Move config migration code into abl-install.sh
- check_lock(): avoid unnecessary file reads
- Avoid unnecessary calls to kill_abl_pids() during action initialization
- Generalize util detection code; simplify and improve logic for util detection
- Detect and check additional essential utils (`cat`, `logger`, `df`, `md5`, `nslookup`, `pidof`, `pgrep`, `netstat`, `ip`). Set and use respective global vars to call these utils.
- Where possible, replace `eval` calls with `export -n` or `local` calls (reduce potential for command injection vulnerabilities)
- When testing adblocking and DNS resolution, use IP addresses specific to each dnsmasq instance, rather than 127.0.0.1. The method `get_dnsmasq_ips()` fetches IP addresses each dnsmasq instance is listening on and stores them in instance-specific context vars.
- Implement smarter way to test adblocking: use blockset-specific test domain by means of calculating MD5 sum of the blockset file and using the resulting string as part of the injected test domain (e.g. `[md5]-adblocklean-test.totallybogus`). This way we can reliably test specific blocksets, and reduce the potential of getting a false positive test result when another adblock-lean instance is active upstream in complex setups. The test domain is appended by the conf-script to the output of the command which prints out the blockset. The logic includes a fallback for the case where required addnmounts are not found and hence conf-script is not possible but we can still load the blockset without conf-script (internally this is referred to as `single-instance mode` because this is exclusively allowed when only one dnsmasq instance is configured): in this case, the domain `[blockset_ID]-adblocklean-test.totallybogus` is appended to the blockset file directly during the final processing stage.

**Config subsystem changes**:

Config is now split between the global config file (`/etc/adblock-lean/global.conf`) and any number of blockset config files (`/etc/adblock-lean/blockset-[blockset_ID].conf`)

Some config options keys have been changed from current master, config migration code should migrate the values automatically.

**Global config:**

Specific methods: `gen_global_config()`, `print_def_cfg_global()`

Options: 

```
PERSIST_HOTPLUG_SCRIPT="0" @ 0|1
blockset_part_failed_action="SKIP" @ SKIP|STOP
min_block_part_entries="1" @ uint
min_ipv4_block_part_entries="1" @ uint
min_allow_part_entries="1" @ uint
max_part_size_KB="${pdc_max_part_size}" @ uint
max_download_retries="3" @ uint
hagezi_default_mirror="github" @ github|gitlab
oisd_default_mirror="oisd" @ oisd|github
stevenblack_default_mirror="github" @ github|sbc_io
deduplication="1" @ 0|1
compression_util="gzip" @ gzip|pigz|zstd|none
unload_blockset_before_update="auto" @ auto|0|1
boot_start_delay_s="30" @ uint
upd_schedule="${upd_schedule:-"0 5 * * *"}" @ string
MAX_PARALLEL_JOBS="auto" @ auto|uint
LOG_VERBOSITY="1" @ 0|1|2|3|4|5
```

**Blockset config:**

Specific methods: `do_gen_blockset_config()`, `print_def_cfg_blockset()`

Options:

```
whitelist_mode="0" @ 0|1
raw_block_lists="${pdc_lists}" @ string
raw_allow_lists="" @ string
raw_ipv4_block_lists="" @ string
dnsmasq_block_lists="" @ string
dnsmasq_allow_lists="" @ string
dnsmasq_ipv4_block_lists="" @ string
hosts_block_lists="" @ string
local_allowlist_path="${ABL_CFG_DIR}/local-allowlist-${set_id}" @ string
local_blocklist_path="${ABL_CFG_DIR}/local-blocklist-${set_id}" @ string
persist_blockset_mode="disable" @ disable|manual|managed
persist_blockset_dir="" @ string
test_domains="google.com microsoft.com amazon.com" @ string
min_good_entries="${pdc_min_lines}" @ uint
max_blockset_file_size_KB="${pdc_max_set_size}" @ uint
custom_script="" @ string
dnsmasq_indexes="${dnsmasq_indexes}" @ uint_list
dnsmasq_conf_dirs="${dnsmasq_conf_dirs}" @ string
```

**Internal API's**:

**Blockset-specific context**:

- `set_params()`, `get_params()` implement storing/fetching blockset-specific params. Params include blockset metadata, blockset-specific config and other runtime parameters.

**Blockset metadata**:

- Mission: store blockset state in a file, retrieve that state at runtime initialization
- Global metadata file path is `/var/run/adblock-lean/blockset-metadata` (volatile) - stores metadata for all [ currently active ] blocksets in UCI format
- Persistent blockset files have individual metadata files stored in the same directory as the persistent blockset file (non-volatile)
- Global metadata params are: `PATH` (path to currently installed blockset file); `SINGLE_INSTANCE` (whether the blockset is only used by one dnsmasq instance, no compression, file placed in the dnsmasq conf-dir, test domain embedded in the blockset file); `MD5` (md5sum of the blockset file); `CNT` (entries count)
- Persistent metadata params are: `PATH`; `MD5`; `CNT`
- Relevant methods: `read_blockset_metadata()`; `commit_metadata()`; `unset_metadata()`

**Initialization**

- Actions specified to the init script are processed in `init_act()`. That method decides which initialization steps are required and executes them.
- As a first step, `init_act()` calls `detect_main_utils()`.
- If traps are required, `init_act()` sets up the traps.
- If lock is required, `init_act()` calls lock-related methods.
- If library scripts are required, `init_act()` calls `source_libs()`.
- If config is required then `init_act()` calls `find_set_configs()`. Next, config (both global and blockset-specific) is loaded via a call to `load_config()`.
- Next, execution is returned to the calling method which implements the user-specified action.
- Global context is initialized via a call to  `set_global_env()`. This sets some system-specific variables, as well as gathers information about dnsmasq instances and the compression utility.
- Blockset-specific contexts are initialized via a call to `set_blocksets_env()`
- `set_blocksets_env()` calls `read_blockset_metadata()`, then calls `get_dnsmasq_ips()`, then (for each individual blockset config) calls `set_bl_env()`.
- `set_bl_env()` sets blockset-specific context, including fetching current blockset run state, checking addnmounts, checking persistent blockset if any, reading persistent blockset metadata and setting parameters required for processing.

**Processing call stack** (simplified):
```
  start()
      |
      V
gen_blocksets()
      |
      V
      |----> gen_set_parts() ----> schedule_jobs() ----> process_set_part()
      |
      V
      |----> gen_blockset()
                       |
                       V
                       |----> print_set_parts()
                       |----> read_list_stats()
                       |----> dnsmasq --test
```

**Debug**:

- Lots of 'normal' debug messages are implemented. Those messages are only printed (via calls to `debug_msg()`) when environment var `ABL_DEBUG` is set.
- Deeper debug is implemented via optional calls to `dbg_start()`, `dbg_end()`. The former essentially runs `set -x` so subsequently all processed commands are printed to console, the latter essentially runs `set +x` to revert that effect. Those methods are never called in the code. To enable deeper debug, manually insert calls to above methods while testing. Additional methods `dbg_on()`, `dbg_off()` mediate this effect to skip printing out spammy sections of the code during deeper debug. Calls to these two methods are already present in some parts of the code.


lynxthecat:
Oh wow, nice. Would this support permanent blocklist on USB stick? Could you outline roughly how you've implemented this?

friendly-bits:
> Oh wow, nice. Would this support permanent blocklist on USB stick?

Yes

> Could you outline roughly how you've implemented this?

The comments in the generated config file expalin how the permanent blocklist feature works. As to nitty-gritty details, there is a lot going on, 2.5k lines of code changed. I've been working on this for a few weeks. So I think it's easier to see the details from the code than to explain them. If you have specific questions, I can answer them.

(normally, I would split this into many small commits, but this implementation required a huge refactor and subsequent bug fixing, so incremental commits didn't make sense to me)

I believe there are some TODO's left in the code yet, so it's not ready to merge quite yet, but at least in my testing everything is already working as intended.

friendly-bits:
Update: there were just 2 TODO's. One is done with above commit. The other one is to parallelize domain lookups - I think  better to postpone this to some later time.

So as far as I'm concerned, this should be good for merging into the `various_improvements` branch and for testing. What do you think @lynxthecat ?

friendly-bits:
I discovered and fixed some minor issues, and made a few additional improvements. Now all seems to be good. Waiting for your input @lynxthecat .

friendly-bits:
^ Some work on improving console messages (mainly spacing) which became a little loose.

lynxthecat:
So the 'managed' option would enable users to free up router memory by storing blocklist entirely on USB? Is it setup such that a restart of DNSMASQ will result in loading in from USB?

friendly-bits:
> So the 'managed' option would enable users to free up router memory by storing blocklist entirely on USB?

Yes



> Is it setup such that a restart of DNSMASQ will result in loading in from USB?

Yes. In managed mode, it will do the processing on the ramdisk and store the final file on the USB. When paused, the backup file will be on the USB as well (assuming that PERM_BLOCKLIST_DIR is on USB). At boot, it will also load the file from USB, rather than downloading and processing a new blocklist. If the specified directory doesn't exist, it will fall back to using the ramdisk.

There is one caveat: if the USB disconnects/is removed after processing is complete and the blocklist is loaded, we have no way to fall back to using the ramdisk on dnsmasq restart. So under these conditions, adblocking will stop working.

friendly-bits:
We could consider implementing a daemon which would run in the background and periodically check whether the blocklist file exists, and restart adblock-lean if it doesn't.

friendly-bits:
Some other things I've been thinking about for the next release:
- Reduce the default value of `boot_start_delay` from 120s to much less, perhaps something like 30s. One frequent request we've heard is to activate adblocking sooner after boot. This seems to be the main reason people want the permanent blocklist functionality implemented as well. Of course, anyone could change the value manually but AFAIK the vast majority of users don't touch the default settings and many are probably not even aware of this option existing. So I'm thinking that reducing this delay will be a welcome change.
- Maybe remove support for the other compression utilities besides gzip, and remove options to call the compression utility with custom arguments. While we have put quite a bit of energy into implementing this support, I strongly doubt that many people are taking advantage of it, and it takes a fairly significant volume of code and, more importantly, increases the maintenance burden (because there are more option combinations to test). I don't feel strongly about this but I would remove these options if I was to decide on this.

lynxthecat:
All those ideas seem excellent to me. I think failing if USB removed is OK. User ought not to expect continued operation then. One question I have is about compression. How does that work with this option?

friendly-bits:
In the meantime, I've come up with an idea of creating a hotplug script which would react on the storage device disconnection and if it's the device the blocklist is stored on, take corrective action. Already did some research and testing on this, seems doable. Considering that USB devices tend to disconnect from time to time, I think this is kind of important to have a reliable fallback.



> One question I have is about compression. How does that work with this option?

I'm not sure what you're asking. Currently we have following options related to compression:

```
# Utility to compress final blocklist, intermediate blocklist parts and the backup blocklist to save memory
# Supported options: gzip, pigz, zstd or 'none' to disable compression
compression_util="gzip"

# Compression options: passed as-is to the compression utility
# Available options depend on the compression utility. '-[n]' universally specifies compression level.
# Busybox gzip ignores any options.
#   Intermediate compression. Default: '-3'.
intermediate_compression_options="-3"
#   Final blocklist compression. Default: '-6'
final_compression_options="-6"
```

There are various bits of code which are implementing all this, e.g. (pointing to the code in this PR):

https://github.com/friendly-bits/adblock-lean/blob/28b45d0be3f369f9accffe3b352c319860021f0a/usr/lib/adblock-lean/abl-process.sh#L51

https://github.com/friendly-bits/adblock-lean/blob/28b45d0be3f369f9accffe3b352c319860021f0a/usr/lib/adblock-lean/abl-process.sh#L719

(if you are asking about how user-specified compression options are applied then the code in the link above is responsible for that)

https://github.com/friendly-bits/adblock-lean/blob/28b45d0be3f369f9accffe3b352c319860021f0a/usr/lib/adblock-lean/abl-process.sh#L94

https://github.com/friendly-bits/adblock-lean/blob/28b45d0be3f369f9accffe3b352c319860021f0a/usr/lib/adblock-lean/abl-process.sh#L122

friendly-bits:
Finally settled on the final implementation and seemingly fixed all bugs. The logic is getting tricky and there are a lot of edge cases. So I had to add a significant amount of "protective code", i.e. all sorts of sanity checks. But at this point I think this is basically release-ready, sans the hotplug idea (which I still want to implement).

lynxthecat:
Regarding compression I meant is the blocklist compression independent of where the blocklist is stored? Any thoughts on whether it is preferable to use compression when using an external storage device? Might it depend on the device (e.g. slow USB stick vs fast USB SDD drive)?

friendly-bits:
> Regarding compression I meant is the blocklist compression independent of where the blocklist is stored?

Yes

> Any thoughts on whether it is preferable to use compression when using an external storage device? Might it depend on the device (e.g. slow USB stick vs fast USB SDD drive)?

IMO it is definitely preferable. The slower the storage medium the more time you gain by using compression. By this logic, compression provides more benefit on a slow USB stick. That said, with file sizes less than 10MiB the time gained would be quite small even on a slow USB drive.

Edit: there is also the benefit of reduced flash wear.

friendly-bits:
> is the blocklist compression independent of where the blocklist is stored

To clarify, it is independent, provided that required addnmounts exist. The new persistent blocklist feature mandates to have an addnmount for the directory specified in PERSIST_BLOCKLIST_DIR. Without this addnmount, adblock-lean will refuse to use or create persistent blocklist. With this addnmount, any files inside PERSIST_BLOCKLIST_DIR are covered by the addnmount and adblock-lean will create a compressed blocklist if required by the config option COMPRESSION_UTIL and if the other required addnmounts (namely `/bin/busybox` and `<compr_util_path>`) exist. If either condition is not satisfied, adblock-lean will fall back to using no compression. For blocklist stored on the ramdisk, the current implementation requires an addnmount for the complete path to the compressed file, e.g. `/var/run/adblock-lean/abl-blocklist.gz` and if that addnmount does not exist then adblock-lean falls back to not compressing the file stored on the ramdisk.

lynxthecat:
Got it - makes sense. Thanks.

Easy for me to write this, but looking over the code modifications, is there any simplifying that could be implemented to reduce the overall complexity of the code? So far, thanks to your attention to detail and thoroughness, adblock-lean has been super robust, but I just want to ensure that there's no simplifying that could be done that might help safeguard that robustness here.

friendly-bits:
The only meaningful simplification I can think of is removing the 'manual' mode for persistent blocklist. Specifically supporting it requires a lot of the logic in order to decide where the new blocklist (and its accompanying "spec", a.k.a. metadata file - see commits description about this) should be written to and to distinguish cases where the code should or should not remove/update the blocklist and the spec file. This would, in fact, reduce the code complexity significantly, but also remove the option of storing the persistent blocklist on the embedded flash. Do you think it should be removed?

lynxthecat:
I think the 'manual' mode makes sense; there could be times when users want to keep their own minimal blocklist for bootup that is not maintained by adblock-lean. Nice feature. So I don't think that should be removed.

friendly-bits:
Your question made me think some more about possible simplifications and I've come up with some ideas, implemented in the commits above. And also some bugfixes and minor tweaks to console messages spacing.

lynxthecat:
Ah fantastic, yes.

friendly-bits:
There is a pause because I'm not entirely satisfied with the implementation, so I'm continuing to work on it.

friendly-bits:
Since this PR requires modification of a substantial portion of logic, it occurred to me that I might as well prepare the code to support multiple blocklists. This adds more work to the current PR but should save a lot of time in the long term. So I'm in the progress of doing that, probably 70% there already. So I'm thinking, if the code is already prepared then why not go all the way and implement support for multiple blocklists already?

As it stands, the main roadblock is the config system, as this requires to have multiple config sections. We would need to either significantly modify the current config system or transition to UCI. In general, I have enough experience with UCI at this point, so it wouldn't be a problem for me to convert the code to utilize UCI, the only thing I don't like about this idea is the fact that UCI config files can not have persistent comments. I.e. as soon as any call to UCI is issued which requires to change anything in the UCI config file, all comments will be stripped from it.

That said, IMO if we want to support multiple blocklists, this is better done with UCI. What's your opinion @lynxthecat ? I'm not yet completely settled on finalizing the multiple blocklists implementation at this PR. Maybe we're better off postponing this and the config code changes.

lynxthecat:
Can you elaborate on what you mean by multiple blocklists? Do you mean facilitating different sets of blocklists and switching between them? Right now we can draw upon multiple blocklist sources (e.g. local and multiple different URLs for remote blocklists). 

Do you suppose adblock-lean will always be OpenWrt only? In cake-autorate we facilitated optional UCI use like this: 

https://github.com/lynxthecat/cake-autorate/blob/8084cca2917bc5c4113f34467e2891bd7998fb9c/launcher.sh.template#L4

I'm uncertain what to think about the UCI aspect myself. In my other projects I've just stuck with config files. I haven't really thought about or used the UCI system enough to be able to make an assessment. So far I'm not altogether clear on the benefits on UCI.

friendly-bits:
> Can you elaborate on what you mean by multiple blocklists?

My idea is to have a notion of "blocklist instance" (I've been struggling to find a good term and just today somehow worked this one out with ChatGPT). So the config file would have multiple sections. One would be "general" with following options:

```
local_allowlist_path="/etc/adblock-lean/allowlist"
local_blocklist_path="/etc/adblock-lean/blocklist"

PERSIST_HOTPLUG_SCRIPT="1" # New option for hotplug script support, not in the PR yet

list_part_failed_action="SKIP"
max_download_retries="3"
deduplication="1"

hagezi_default_mirror="github"
oisd_default_mirror="oisd"
stevenblack_default_mirror="github"

# Following three options are probably better off in the "general" section, just to make the config simpler.
min_blocklist_part_line_count="1"
min_ipv4_blocklist_part_line_count="1"
min_allowlist_part_line_count="1"

# We should probably remove following compression-related options as discussed above, but including them for now
compression_util="gzip"
intermediate_compression_options="-3"
final_compression_options="-6"

unload_blocklist_before_update="auto"
boot_start_delay_s="30"
MAX_PARALLEL_JOBS="auto"
custom_script=""
cron_schedule="0 5 * * *"
LOG_VERBOSITY="1" # new option introduced in this PR

```

And then there would be 1 or more sections, each one defining a "blocklist instance", with following options in each:

```
whitelist_mode="0"
raw_block_lists="hagezi:pro.mini"
raw_allow_lists=""
raw_ipv4_block_lists=""
dnsmasq_block_lists=""
dnsmasq_allow_lists=""
dnsmasq_ipv4_block_lists=""
hosts_block_lists=""
PERSIST_BLOCKLIST_MODE="manual"
PERSIST_BLOCKLIST_DIR="/mnt/test/abl"
test_domains="google.com microsoft.com amazon.com"
min_good_line_count="24000"
max_file_part_size_KB="2000"
max_blocklist_file_size_KB="2000"
DNSMASQ_INDEXES="0"
DNSMASQ_CONF_DIRS="/tmp/dnsmasq.cfg01411c.d"
```

The user would set different DNSMASQ_INDEXES in each section and this way, they could have different adblocking/content blocking for separate network segments. In addition, we could look into implementing support for mac-based DNS requests redirection to a specifc dnsmasq instance - this would allow the user to have more surgical control over client devices (similar to what Dirk recently implemented for 'adblock').

friendly-bits:
> Do you suppose adblock-lean will always be OpenWrt only?

Currently we are relying heavily on OpenWrt-specific features. For that matter, the init script is specific to OpenWrt. The shell language we are using is specific to Busybox ash. The code should work fine in Bash as well, but not in other shells like dash. I also just rewrote the code which keeps track of blocklist metadata - the current implementation utilizes UCI for this purpose. This is not a huge chunk of code but the format requires config file with sections (to support separate metadata for different blocklist instances). Reimplementing this in shell would constitute a pretty big chunk of work. Same applies to our main config file. We are currently using a custom config system with custom code, and in order to support section, a lot of that would have to be rewritten and the code would probably become much more complex. This is why I'm thinking that UCI would be a better fit, as UCI already supports multiple sections out-of-the-box. While I really like our current config system, objectively I think that for a more complex config, it would be better to transition to UCI.

lynxthecat:
Ah yes, multiple dnsmasq instances. 

Couldn't that support, for example, routing child-safe traffic to one dnsmasq instance, and the remaining traffic to another dnsmasq instance? I think we've exchanged thoughts on this before at one time. You identified a possible issue relating to setting up the dnsmasq instances. I'm sure there'd be a solution of one form or another.

Or zero downtime by priming a new instance and switching over rather than the delay we presently have. So I mean that we use existing instance whilst new instance is getting loaded. Then when new instance is loaded we switch over.

Or of course just dealing with separate network segments each having their own dnsmasq instances.

friendly-bits:
> multiple dnsmasq instances.

Multiple dnsmasq instances are already supported in current adblock-lean. We do currently have a limitation, as any configured dnsmasq instances will use the same blocklist. The multiple **blocklist instances** idea intends to solve that. Each such blocklist instance could be configured to be loaded by any (or all) dnsmasq instances.



> routing child-safe traffic to one dnsmasq instance, and the remaining traffic to another dnsmasq instance?

I'm not sure I understand. How would dnsmasq know which traffic is child-safe? If you would want to have separate content blocking for children, multiple blocklist instances would allow for that. E.g. you set up a separate network segments for grownups and for children (perhaps VLANs), you configure 2 dnsmasq instances. For one dnsmasq instance, you specify one network interface (corresponding to one VLAN), for the other dnsmasq instance you specify another network interface (corresponding to the other VLAN). Then you configure 2 blocklist instances in adblock-lean, and you specify a different dnsmasq instance index for each. Then you have separate content blocking.


> I think we've exchanged thoughts on this before at one time. You identified a possible issue relating to setting up the dnsmasq instances.

I believe you are referring to our discussion concerning the other idea you mentioned: zero downtime. This I don't think is technically possible because only one dnsmasq instance can bind to a particular network interface and port at a time, and because AFAIK there is no way to reconfigure a dnsmasq instance on-the-fly to bind to a different network interface and port. Essentially, you tell dnsmasq which network interface and port to bind to when you call dnsmasq. Which makes it impossible to have one instance serving clients and another one loading blocklist and later on replacing the first one. The second instance will simply refuse to run.

friendly-bits:
For reference, here is the config system I implemented for QosMate. The format is UCI but I wrote a custom parser in awk which validates the config (like our config management system does in adblock-lean) and generates shell code which is then `eval`'d to have the variables set. I went a long way to sanitize the entries, so `eval` is probably safe there, but actually there is a useful trick I learned from scipts shipped with OpenWrt: instead of `eval`, we could use the ~~`import`~~ (Edit: `export`) command. This makes it completely safe.

https://github.com/hudra0/qosmate/blob/25e2903086047edfaebb03d8be4635e326f5e053/etc/init.d/qosmate#L1194

The following function calls `parse_config()`, `eval`'s the result and handles errors if any:

https://github.com/hudra0/qosmate/blob/25e2903086047edfaebb03d8be4635e326f5e053/etc/init.d/qosmate#L1312

lynxthecat:
> If you would want to have separate content blocking for children, multiple blocklist instances would allow for that. E.g. you set up a separate network segments for grownups and for children (perhaps VLANs), you configure 2 dnsmasq instances. For one dnsmasq instance, you specify one network interface (corresponding to one VLAN), for the other dnsmasq instance you specify another network interface (corresponding to the other VLAN). Then you configure 2 blocklist instances in adblock-lean, and you specify a different dnsmasq instance index for each. Then you have separate content blocking.

Yes this is the sort of thing I had in mind. 

> I believe you are referring to our discussion concerning the other idea you mentioned: zero downtime. This I don't think is technically possible because only one dnsmasq instance can bind to a particular network interface and port at a time, and because AFAIK there is no way to reconfigure a dnsmasq instance on-the-fly to bind to a different network interface and port. Essentially, you tell dnsmasq which network interface and port to bind to when you call dnsmasq. Which makes it impossible to have one instance serving clients and another one loading blocklist and later on replacing the first one. The second instance will simply refuse to run.

I'm certain there's a way round this one way or another. This is perhaps an extreme example to give, but  when setting up cake instances it is routine to create a dedicated IFB interface on which to set up a cake interface (e.g. to deal with ingress traffic) - see e.g. [here](https://github.com/lynxthecat/cake-qos-simple/blob/ac8fdeeed96639d2df1812af85647da690a22086/cake-qos-simple#L255), and so analogously I imagine it would be possible to set up an IFB interface and set up the further dnsmasq instance on that. I'm just opining that I think there is a way to implement this zero-downtime switchover concept. Whether the complexity and processing cost is worth it I'm not sure - especially for the typical home network use case when everyone is asleep in the middle of the night. On the other hand, surely there are situations where you want zero downtime in respect of filtered DNS queries (if only because there's plenty of memory and processing present and one wants to maximise filtered DNS performance)?

friendly-bits:
I pushed my current work-in-progress with the commits above ^.

This already has most commands working correctly (start, stop, pause, resume, setup and others) but `status` is not yet working properly. Moreover, support for multiple blocklists will require config implementation changes which are currently not implemented and for now, there is an ugly hack in the code just to allow testing. Hotplug script implementation is unfinished and currently disabled in the code. There are lots of TODO's in the code to work through. And there are surely lots of bugs in there as well.

friendly-bits:
Above commits mostly complete the implementation of support for multiple blocklists, including significant changes to the config system. There are still some TODO's left in the code, and migration from the previous config system is not yet implemented.

friendly-bits:
^ Above huge commit introduces many code changes, including:
- Use the term `blockset` when referring to the final blocklist, in order to avoid confusion between that and blockset part of type `blocklist`.
- Separate fetch process from blockset installation process: now all list parts are fetched first, and at a second stage out of those fetched parts any required blocksets are constructed
- Reliably keep track of blockset state during runtime - this allows to avoid unnecessary ("just in case") execution of commands, including costly dnsmasq restarts
- Move config migration code into the abl-install script
- Improve console messages
- Fix numerous edge case bugs

friendly-bits:
Updated the first comment with rough outline of the changes.

himanshu-mahendra-jain:
@friendly-bits @lynxthecat When is the release? Is more testing needed? Thanks for the great work!

